PR #7, opened 2023-04-11
src/confd: add the ability to change MotD in cli

rical:
# Pull Request
Not sure if this is even something we want? I did it mainly to learn and to test the work flow.
If this is indeed something we want then I intent to fix the duplication of the default value (in skel /etc/motd and in the confd C code).
I might also hook it into the `aug` framework, even tho it should never fail (TM) and "abort" seems to work as intended?

# Commit message
Add the ability to set and delete system MotD (Message Of The Day) using the cli.

So there's no staging concept (aug), if user issues "leave" the change takes immediate effect.

The default value is hard coded in two places, confd and the rootfs file /etc/motd. The intention is to fix this in upcoming patches.

Introduces:
exec> configure> set system motd STRING
exec> configure> delete system motd

### DIFF
--- a/package/sysrepo-plugin-system/sysrepo-plugin-system.mk
+++ b/package/sysrepo-plugin-system/sysrepo-plugin-system.mk
@@ -22,6 +22,8 @@ define SYSREPO_PLUGIN_SYSTEM_INSTALL_YANG_MODELS
                 $(TARGET_DIR)/usr/share/yang/modules/sysrepo-plugin-system/iana-crypt-hash@2014-08-06.yang
 	$(INSTALL) -D -m 0644 $(@D)/yang/ietf-system@2014-08-06.yang \
                 $(TARGET_DIR)/usr/share/yang/modules/sysrepo-plugin-system/ietf-system@2014-08-06.yang
+	$(INSTALL) -D -m 0644 $(@D)/yang/infix-system@2014-08-06.yang \
+                $(TARGET_DIR)/usr/share/yang/modules/sysrepo-plugin-system/infix-system@2014-08-06.yang
 endef
 SYSREPO_PLUGIN_SYSTEM_POST_INSTALL_TARGET_HOOKS += SYSREPO_PLUGIN_SYSTEM_INSTALL_YANG_MODELS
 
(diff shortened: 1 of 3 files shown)


troglobit:
First of all, Yes, this is something we want, and it's a great example to start with I think!

Now, there are a few things that I believe need to be fixed:

  1. The standard models must be kept as-is, extensions should be in separate yang file(s) that we load in (in this case) `ietf-system.c` and use the augment syntax.  See `ietf-ip.yang` for an example.  Also, even though we've decided to go with IETF, there's another world of YANG models there that might serve as inspiration https://github.com/openconfig/public/blob/master/release/models/system/openconfig-system.yang (syntax, description, and types used for validation come to mind.)
  2. You should add `switch (event)` in your callback so you don't regenerate the file multiple times (you'll get at least two events per change), this will also allow you to handle abort without too much effort, or like I did with `change_hostname()`, just ignore all events except DONE (last event) and just trust that recreating the file(s) work and validation is done by libyang.

Oh, and you don't need to remove the duplicate /etc/motd file.  Infix has three operating modes (selectable with the `setup` tool) and one is the native mode in which users edit /etc manually. These modes share the same rootfs.  Your default MOTD can be something like "To enter the CLI, type 'cli' at the shell prompt."  or "Remember to change your default password at first login.", something more geared towards NETCONF (operating mode) users.

Last, Augeas is a really cool project and for anything more complex it's super useful. However, it's perfectly OK to render all simple files manually like you've done (and I did with chrony.conf).  After all, we've got a lot of models to implement ... ;-)

troglobit (commented):
Looks great, we just need to rebase+squash at merge.  Handing over to @wkz for final review.

src/confd/yang/ietf-system@2014-08-06.yang
```diff
@@ -234,6 +234,11 @@ module ietf-system {
        "The name of the host.  This name can be a single domain
         label or the fully qualified domain name of the host.";
     }
+    leaf motd {
+      type string;
+     description
+       "Set the MotD (Message of the Day), which is shown after login";
+    }
```
troglobit: Please don't modify std models, squash with last commit.

troglobit (changes requested):

wkz (approved):

PR #8, opened 2023-04-13
Sysrepo build fixes

wkz:


troglobit (approved):
Awesome, a great cleanup that also answered a few style questions I've had.  Thanks for doing this!

PR #11, opened 2023-04-13
Housekeeping cleanup

troglobit:
- Drop unused packages from tree:
  - sysrepo-plugin* + umgmt
  - ifupdown2 + deps
  - clixon + cligen + mech
- Update README, we're using sysrepo now
- Drop unused packages from defconfigs:
  - ttyd
  - tmux
  - mrpd
  - mrp


wkz (approved):
Kill it with :fire:!

PR #14, opened 2023-04-19
x86_64: Harmonize GRUB's behavior with that of U-Boot

wkz:
This means that:

- The full boot order is now respected, using the magic "fallback" varible in GRUB. This means that if an entry fails to load, we move on to the next one in the defined order.

- Netbooting is supported. In this mode, GRUB will acquire an IP address using DHCP, after which it downloads an image using TFTP.

- If we run out of valid entries, we always make sure to reboot. This ensures that, e.g., a temporary TFTP server outage does not cause an Infix device to get stuck in GRUB.

At the moment, the image handling is excruciatingly inefficient when netbooting, as the image is downloaded multiple times behind the scenes. Ideally, downloads would be cached, but a quick investigation indicates that an implementation of this would take more time than we can spare at the moment. Therefore, we settle for the working, though wasteful, implementation.


troglobit (approved):
lgtm, nice work!

PR #22, opened 2023-05-10
confd: add global event hook support

troglobit:
Refactor `register_change()` wrapper, for `sr_module_change_subscribe()`, with an additional twist of setting up a commit DONE event catcher.

This creates a global event hook that ensures we call `initctl reload` after all module callbacks have completed, regardless of the xpath and without having to have "shadow" subscriptions to all new modules being added to confd.

Based on on ideas from https://github.com/sysrepo/sysrepo/issues/2188

### DIFF
--- a/src/confd/src/core.c
+++ b/src/confd/src/core.c
@@ -4,6 +4,10 @@
 
 static struct confd confd;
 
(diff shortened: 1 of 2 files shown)


troglobit:
Closing in favor of a clean new PR.

PR #25, opened 2023-05-12
wkz/confd: Basic ietf-{interfaces,ip} support

wkz:
Add support for:
- Enabling/disabling interfaces
- Adding/removing IPv4/6 addresses
- Enabling/disabling IPv4/6 autoconf
- Enabling/disabling IPv4/6 forwarding


troglobit (approved):
Neat and tidy, a very much needed cleanup!  No comments, looks great!

PR #33, opened 2023-05-23
Initial support for DHCPv4 yang model

troglobit:


troglobit:
I have several questions on how to proceed with this, most importantly:

 - should we care about DHCP options at all for this first prototype, or just go with the defaults encoded in udhcpc?
 - should I spend some time to get basic DHCPv6 client support as well?

troglobit:
We agreed AFK that:

 - DHCP options could be skipped for now (will be a yang revision later)
 - DHCPv6 is not required yet, and not as load bearing as DHCP on IPv4
 - Basic DHCPv4 client-id is, however, needed now.

troglobit:
@wkz review issues should be solved now, root cause was `initctl` waiting for y/N? input when deleting the configuration file.

PR #37, opened 2023-05-30
Add support for VLAN sub-interfaces

troglobit:
Basic support for VLAN sub-interfaces:

```
  interface eth0.10 {
    type ethSubInterface;
    encapsulation {
      dot1q-vlan {
        outer-tag {
          tag-type s-vlan;
          vlan-id 10;
        }
      }
    }
    parent-interface eth0;
    ipv6 {
      autoconf {
        create-global-addresses true;
      }
    }
  }
```

### DIFF
--- a/board/common/rootfs/sbin/dagger
+++ b/board/common/rootfs/sbin/dagger
@@ -78,7 +78,8 @@ do_abandon()
     [ "$next" ] && [ -d "$basedir/$next" ] \
 	|| abort "Next generation does not exist"
 
-    rm -rf "$basedir/$next" "$basedir/next"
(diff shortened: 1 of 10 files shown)


wkz:
@troglobit : I tweaked some stuff :sweat_smile: 

Please have a look/test drive and see if it works for you.

troglobit:
> @troglobit : I tweaked some stuff :sweat_smile: 
> 
> 
> 
> Please have a look/test drive and see if it works for you.

Nice work! I have some questions, but we can take them IRL. Also, for some reason rebuilding with these changes fails to boot my system. I'll dig into it some more and get back to you.

PR #39, opened 2023-06-02
Export qemu.sh and qemu/Config.in to release tarball

troglobit:
This changeset makes it possible for a user to easily test their Infix build in Qemu.

The qemu.sh script has been extended with a `-c` flag to start the kconfig-frontends tool `kconfig-mconf` with a slightly modified qemu/Config.in 

To test: 

     sudo apt install kconfig-frontends

### DIFF
--- a/board/common/post-image.sh
+++ b/board/common/post-image.sh
@@ -52,3 +52,7 @@ fi
 
 # For use outside of the build system, e.g., Qeneth
 ln -sf rootfs.squashfs "$BINARIES_DIR/$NAME$(ver).img"
+
+# Create the necessary files to run Qemu from a build
+grep QEMU_ "$BR2_CONFIG" | sed 's/"images[\/]*/"/g' > "/$BINARIES_DIR/qemu.cfg"
+cp "$BR2_EXTERNAL_INFIX_PATH/qemu/qemu.sh" "/$BINARIES_DIR/"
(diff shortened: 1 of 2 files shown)


board/common/post-image.sh
```diff
@@ -52,3 +52,9 @@ fi
 
 # For use outside of the build system, e.g., Qeneth
 ln -sf rootfs.squashfs "$BINARIES_DIR/$NAME$(ver).img"
+
+# Create the necessary files to run Qemu from a build
+grep QEMU_ "$BR2_CONFIG" | sed 's/"images[\/]*/"/g' > "/$BINARIES_DIR/qemu.cfg"
+sed 's/"images[\/]*/"/g
+     s/comment "System setup"/mainmenu "Qemu Settings"/g' < "$BR2_EXTERNAL_INFIX_PATH/qemu/Config.in" > "/$BINARIES_DIR/Config.in"
+cp "$BR2_EXTERNAL_INFIX_PATH/qemu/qemu.sh" "/$BINARIES_DIR/"
```
wkz, line 60: Is there any reason to keep the old config interface (i.e. via Buildroot's menuconfig) around? Maybe we could just
- Move `/qemu` to `/board/common`
- Let `post-image` install them in `$O/images`
- Have a convenience `make run-menuconfig` target in `external.mk`
 
That should get us out of the need for `sed`s, and it ought to make the logic in `qemu.sh` simpler as well.

troglobit: Not really, for me this was more or less just the first step. But you're right, no point in keeping the old interface, let's steam ahead!

I'll push some fixes during the weekend along the lines of your proposal.

troglobit: There, fixed in 0154c0f :sunglasses:

PR #41, opened 2023-06-05
confd: move src into confd subdirectory

rical:
This patch is a preparation for the upcoming "statd" addition.


troglobit:
lgtm, @wkz what do you think?

PR #42, opened 2023-06-07
Add new statd skeleton for operational status

rical:
Add a new "statd" module to sysrepo. This module is intended to handle
operational status, i.e. R/O attributes.

In this state, the statd module only retrieves and answers to the
"oper-state" of "eth0". This will be replaced by proper interface
handling and json data processing in upcoming patches. The indent here
is to get the base logistics in place as soon as possible to make life
easier for developers.


troglobit:
LGTM. @wkz ?

troglobit (approved):

PR #43, opened 2023-06-07
Test suite

wkz:


troglobit:
Btw, I noticed the first run failed due to insufficient access, so I added `infix` to the list here https://github.com/orgs/kernelkit/packages/container/infix-test/settings

troglobit:
> I ended up making the package public (which it should be anyway) so that works now. The last fail was due to KVM not being available. So the build running now should be able to handle that.

Awesome, yeah that's a good idea! :-)

> Are you able to run the tests locally? (`make infix-check`)

I've just sat down again after weekly customer followup, so will get right on playing with it! Looks great, as per usual, code wise.

troglobit:
Works. No risk of regressions, merging.

PR #50, opened 2023-06-13
statd: add limited operational data for all interfaces

rical:
Add oper-status and if-index data to sysrepo for every existing
interface. The origin of the data is the "standard" ip (link) tool with
JSON output.

Current implementation is limited to interfaces existing at startup.


troglobit (approved):
Tests pass, and everything looks good. So merge! :-)

Just a side note: you've added an `openlog("statd")` that changes the log output also for confd, we can have a talk about that AFK how to deal with that.

PR #52, opened 2023-06-15
Infix doc mgmt

jovatn:
Adding rudimentary information on discovery of Infix units and Infix Discovery services.
Is title ok?
Should we add links/references or more info on mDNS/DNS-SD or SSDP? I have not even spelled out the acronyms ...

But hopefully this is a good start. :-)

### DIFF
--- a/doc/discovery.md
+++ b/doc/discovery.md
@@ -1,0 +1,193 @@
+Discover Infix Units
+====================
+
+```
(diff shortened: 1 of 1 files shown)


troglobit (approved):
lgtm, we can add explanations of acronyms later.  Nice work, as always! :)

PR #53, opened 2023-06-16
Initial bridge and veth support

troglobit:
This series adds support for a basic (non-vlan filtering) bridge and veth pairs.  During the course of these changes I discovered that we didn't honor the ipv4/ipv6 enabled node, so that was added.  This in turn led to the discovery of our factory defaults not being properly set in sysrepo, which led to upstream report sysrepo/sysrepo#3079.

The new sysrepo bootstrap is handled by `src/confd/sysrepo-bootstrap.sh`, which means all modules required by our plugins have also been moved here so the system factory-default is properly initialized in all modules.


troglobit:
Tests continue to fail on Azure, while still working perfectly locally, even if I download the resulting artifacts and run the tests on them!  We need to have a talk about this because I've basically run out of ideas.

> NOTE: lots of desperate and brain dead commits now on this branch, so I've converted it to a draft to indicate it needs to be cleaned up before we merge it!

troglobit:
@wkz the `ieee-group-forward` enum was a bit hard to parse, but now it works!  The rest of the commits have also been cleaned up a bit and should be ready for a review, finally :-)

troglobit:
I've just pushed a fix for the `RpcError` exception bug I introduced, but otherwise this should be good to go.  There's lots of room for improvement; removal of debug logs, moving bridge port setup from ports to the bridge, etc.  I'll take care of it in the next branch, `bridge-vlans` that I'm already working on.

src/confd/src/confd/ietf-interfaces.c
```diff
@@ -541,24 +728,53 @@ static int netdag_gen_afspec_set(struct dagger *net, struct lyd_node *dif,
...
+	struct lyd_node *node;
+	int err = 0;
+
+	node = lydx_get_descendant(lyd_child(cif), "bridge-port", NULL);
+	if (node) {
+		const char *brname = lydx_get_cattr(node, "bridge");
+
+		fprintf(ip, " master %s", brname);
+
+		err = dagger_add_dep(net, ifname, brname);
```
wkz, line 755: Isn't this still inverted? The bridge should depend on all its ports being setup before running its own setup?

I think the `link set dev PORT master BR` needs to go in an `.ip` file in the _bridge's_ action directory (at a priority that runs after the bridge itself is created, but before applying VLANs etc.

Also, the fact that `ietf-interfaces.c` does not contains the string `nomaster` has me worried that we're not handling the case when interfaces are removed from a bridge? Correspondingly, that command should also run in the _bridge's_ exit action, I think.

wkz (commented):
Awesome job on this! So nice to have a proper factory config too! :heart: 

There are some questions remaining around dependencies etc (see my comment). I'm happy to merge this though, then we can improve things from `main`. What do you think? I can take a crack at the dep-stuff if you've (understandably) had enough of this code for a while :smile:

PR #55, opened 2023-06-22
Add NETCONF support for Linux bridge native VLANs

troglobit:
Example:

```
  interface br0 {
    type bridge;
    enabled true;
    bridge {
      vlans {
        pvid 4094;
        vlan 10 {
          untagged e2;
          tagged br0;
        }
      }
    }
  }
```


wkz:
:fire:!

PR #63, opened 2023-06-28
confd: fix memory leak in lydx_new_path()

rical:
# Disclaimer 
This might break something. **Please test your use case**.
I didn't find any need for the "first" concept at all. I have tested to query the operational status for both ietf-system (confd) and ietf-interfaces (statd). Sysrepo is still able to report operational status in the same manner as before and the memory leak is fixed.

Here's what I used to test this:
```sysrepocfg -X -d operational -m ietf-system```
```sysrepocfg -X -d operational -m ietf-interfaces```

# Patch
Prior to this patch, every new call to lydx_new_path() where "first" was set to 1 resulted in a memory leak of 72 bytes. Typically at least once for each sysrepo query.

The memory was allocated deep down in the lyd library, but as the parent node wasn't set (NULL) it wasn't found by sysrepo when cleaning up.

**In this patch we mitigate this by simply removing the "first" concept, as it doesn't appear to be needed.**

### DIFF
--- a/src/confd/src/confd/ietf-system.c
+++ b/src/confd/src/confd/ietf-system.c
@@ -138,7 +138,6 @@ static int clock_cb(sr_session_ctx_t *session, uint32_t sub_id, const char *modu
 	const struct ly_ctx *ctx;
 	char curtime[64];
 	time_t now, boot;
-	int first = 1;
 	int rc;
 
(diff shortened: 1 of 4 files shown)


troglobit:
Awesome work on tracking this leak down! LGTM and tests pass.

PR #71, opened 2023-07-02
klishd: drop privileges in client connection

troglobit:
When a client connects over the AF_UNIX socket we should drop privileges to match the uid and gid of that user.  The '%u' translation done by the klish_prompt() function needs the USER env. variable to be set, so we do that too here.


troglobit:
This was a good idea.  Letting the forked-off child process of klishd drop its privileges so it connects as, e.g., the `admin` user.  Unfortunately it does not work.

Later revisions of the klish project add their own UID tracking, but the daemon still runs and operates as root when connecting to sysrepo.  

In PR #90 I explored this further and found several possible reasons for the problems I saw.  These are documented in the patch on our kkit branch of klish, which we switch to in PR #90.  Since that supersedes this PR, I suggest we close this one instead.

Issue #74, opened 2023-07-03
Drop clean-etc script


troglobit:
For everyone's sanity, we should drop the clean-etc script installed by confd.  I propose reducing the amount of services enabled by default to an absolute minimum that can be shared between the different Infix flavors, and that the _Classic_ flavor instead maintain a set of services enabled by default in `board/classic/rootfs`.

PR #75, opened 2023-07-03
Misc fixes, part 2

troglobit:
The motivator for this PR was issue #74, which we both ran into hard today.  But since I'm in cleaning mode, I've not only dropped the `clean-etc` script, but also went through the Finit .conf files that are enabled by default, and adjusted accordingly.  A few more services now run, e.g. SNMP, which we want anyway but don't have modeled in YANG yet.

As a bonus, I noticed and fixed some packaging issues with missing LICENSE statements (required by our customers).

A recurring extra fun-fun thing is the long running effort at addressing Bob's reluctance to upload latest built artifacts to our Latest Build release.  This one will be the death of me.


rical (approved):
LGTM, with the disclaimer that I don't know the reasoning behind etc-clean and why it was created. But I agree it caused some headache.

PR #76, opened 2023-07-03
Updating deviations for ietf-system.yang (rw-part)

jovatn:
Proposed updates for deviations
- timezone-utc-offset: cutting out more of that 'branch'
- radius: not supported
- dns-resolver udp/tcp-port: not configurable

Also, I do not know how to do with "ro-part" and "rpc-part". I guess this is not yet implemented, or at least not available through the CLI. infix-system-software is adding infix-specific stuff, thus I do not wish to touch ro or rpc.


troglobit:
Interesting, didn't even consider setting radius as a deviation since it's a feature we don't declare. (Feature guarded parts of a model are listed in tree mode as **{radius}?**)  Wonder what's the recommended way of handling that?  My guess is that a client should ask for a feature list as well as the models when initiating a connection.

The other stuff looks good, but like you say, the revision needs to be stepped, and then you also need to update the corresponding C file (src/confd/ietf-system.c is where I believe it's located currently).

Regarding ro-part and rpcs, we support all of them :sunglasses::v: so no need for deviations there.

jovatn:
Thanks for quick feedback. Let's chat a bit tomorrow. 
Although it makes sense for clients to ask for feature list, humans can't.
I mean, we can not use the deviations' list as standalone info of what is supported. 
Another good thing with explicitly stating missing optional features is that its easy to keep track when its supported - the developer would need to update/remove the deviation. :-) 

This said, I do not claim to know the correct way to use deviations. You're remark is valid.
In addition, the "dns-resolver udp/tcp-port" feature which I put a deviation for is also optional to support ...

troglobit:
Yeah I agree, it really isn't clear what the best way of handling all this really is. I'm just slightly concerned that having two "locks" (feature flag + deviation in another file) could be confusing.

Looking forward to hearing what you've learned tomorrow! :smiley:

troglobit:
Your branch now has conflicts with `main`, in a file I don't think you've edited: `.github/workflows/build.yml`

Try `git pull --rebase origin/main` or similar to put your own commits on top of the latest.

troglobit:
Absolutely brilliant stuff!  Really looking forward to discuss this with @wkz when he gets back from his vacation.  

Nice work, @jovatn! :clap::clap::clap:

troglobit:
This looks great, and the tests pass, so after careful consideration I've decided to merge it!  Infix is a lot better off with this in place rather than without.  We can still have a chat with @wkz about this, and if we change our minds we can revert parts of this later.

PR #79, opened 2023-07-04
test/case: Add authentication/user test

ahmkar94:
Added a test case for adding/deleting a user

### DIFF
--- a/test/case/ietf_system/add_delete_user.py
+++ b/test/case/ietf_system/add_delete_user.py
@@ -1,0 +1,72 @@
+#!/usr/bin/env python3
+
+import infamy
+import copy
+import crypt
+import random
+import string
+import re
+
+def generate_restrictred_credential():
+    credential = "".join(random.choices(string.ascii_lowercase, k=64))
+
+    while not re.match("^(?![0-9])[\w]+$", credential):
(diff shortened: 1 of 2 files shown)


troglobit (changes requested):
Just a small request, in the steps after "Add new user", maybe add user/pass in the `with` lines, like Tobias did in the hostname test?

     with test.step(f"Verify new hostname ({new})"):

This makes it more obvious that we actually randomize the username and password.

Other than that, it looks good, nice job!  Let's continue like this, one PR per test.   When you run out of these obvious test cases then we can have a talk about the next round before we have a real sitdown about more comprehensive use-case tests :-)

ahmkar94:
I see, I did have those in the step name until we decided to use 64 characters. 
Would it be ugly to see random 128 characters in the test-step-name? 
We have 3 options: 
- To show only username
- To reduce number of characters
- To let the title be long and ugly :-D

image

troglobit:
Hmm, good questions!

> Would it be ugly to see random 128 characters in the test-step-name? We have 3 options:
>     * To show only username

Nah

>     * To reduce number of characters

How about adding random length between 8-64 chars?

>     * To let the title be long and ugly :-D

That's not too bad, I say we let it be long :)

troglobit (approved):

PR #80, opened 2023-07-04
New standalone status daemon

rical:
Break out operational status into new daemon

This PR does two main things:
1) Breaks "statd" free from "confd".
2) Introduces netlink monitoring of interfaces for sysrepo data.

Prior to this PR, sysrepo operation status was handled by code
named "statd" inside "confd".

In this PR, we break out the interface part of the operational
sysrepo status, previously "statd", into a new daemon, also called
"statd".

The main reason for doing this is to avoid sysrepo threading, which
allows the new "statd" daemion to have control over sysrepo events in the
same event pool as netlink events from the kernel. This in turn allows
statd to subscribe and unsubscribe dynamically to interfaces that
comes and goes during runtime (such as vlans).

It also makes sense that "statd" is in fact a finit daemon.


troglobit:
Looks great, nice work on this!

troglobit (approved):

PR #83, opened 2023-07-06
Authentication improvements

troglobit:
- Update klish-plugin-sysrepo to get EnableNACM support so we can edit nacm rules in CLI
- Add ietf-system augment for specifying login shell, default `/bin/clish`
- Allow adding remote-only users that are not allowed password login, only SSH keys
- Add compat symlink to ~/.ssh/authorized_keys for netopeer2-server
- Misc fixes.


src/confd/src/confd/ietf-system.c
```diff
@@ -579,10 +582,39 @@ static int is_valid_username(const char *user)
...
+
+		if (!lyd_find_path(cfg->tree, xpath, 0, &node))
+			shell = (char *)lyd_get_value(node);
+	}
+
+	/* Verify the given shell exists (and is a login shell) */
+	if (!shell)
+		shell = LOGIN_SHELL;
+	else if (!whichp(shell))
+		shell = LOGIN_SHELL;
```
wkz, line 603: Won't this still allow me to set `/usr/bin/mg`, for example, as my shell? Which is then disallowed by `/etc/shells` when I attempt to login?

src/confd/yang/infix-system@2023-07-06.yang
wkz, line 1: Should we prune the old version from the source tree?

src/confd/src/confd/ietf-system.c
troglobit: Yeah, you're right.  We should probably add something like https://github.com/kernelkit/infix/commit/76c09ba3bdf4d94a11b1bc5bcdd24c396f3d7863 to the model instead, and this passage could instead be reduced to falling back to `/bin/false`.

src/confd/yang/infix-system@2023-07-06.yang
troglobit: Pending your return from vacation, me and @jovatn had a discussion about this and decided in the meantime to keep older versions.  Can't for the life of me remember why atm., but maybe he remembers?

wkz: Alright, we can always fix that later on :+1:

src/confd/src/confd/ietf-system.c
wkz: We'll leave it as-is for now, and track this issue separately in #95.

troglobit:
Thank you for taking the time up fix the conflicts! I'll have a look at #95 when I get back from vacation.

PR #84, opened 2023-07-06
Doc: Fixing hybrid-mode FRR example

jovatn:
two fixes
- first a typo (probably copy/paste issue)
- the sleep and 'vtysh -b' was added to make vtysh read the /etc/frr/frr.conf file upon start(*) sleep needed to make this happen after initctl reload

(*) See https://docs.frrouting.org/en/latest/vtysh.html#integrated-configuration-mode But there ought to be a better way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,9 +70,9 @@ At bootstrap Finit can start user scripts from a [run-parts(8)][] like
 directory: `/cfg/start.d`.  For example, the following starts OSPF:
 
 ```sh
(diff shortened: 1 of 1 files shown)


jovatn:
I happend to push to my forked main branch rather to my 'infix-readme' branch. I can redo the PR correctly if that causes issues.

troglobit:
May cause you some issues in your fork, but it looks good here, so merge! :-)

PR #85, opened 2023-07-07
linux: add loopback carrier for operstate up

troglobit:
This is a proposal to fix the issue of loopback having operstate UNKNOWN, which otherwise propagates over NETCONF as well.

See separate discussion thread in Projects.

### DIFF
--- a/src/confd/src/confd/ietf-interfaces.c
+++ b/src/confd/src/confd/ietf-interfaces.c
@@ -1021,7 +1021,7 @@ static sr_error_t netdag_gen_iface(struct dagger *net,
 	/* Bring interface back up, if enabled */
 	attr = lydx_get_cattr(cif, "enabled");
 	if (!attr || !strcmp(attr, "true"))
-		fprintf(ip, "link set dev %s up\n", ifname);
+		fprintf(ip, "link set dev %s up state up\n", ifname);
 
 	err = err ? : netdag_gen_sysctl(net, dif);
 


wkz:
Supplying a known operstate when we bring up lo should be sufficient to not have to carry this kernel patch.

I.e. doing the equivalent of `ip link set dev lo state up` accomplishes the same thing, me thinks. NOTE: This is (confusingly) not the same thing as `ip link set dev lo up`. The former sets `IFF_UP` in the device flags, while the latter sends an updated `IFLA_OPERSTATE`.

troglobit:
Awesome!  And yeah, you're right of course, just verified locally.  Unsurprisingly this command isn't documented in `ip link help` nor anywhere in the manpage `ip-link(8)`, which explains why I didn't find it while looking.  

So something like this, right?

```diff
diff --git a/src/confd/src/confd/ietf-interfaces.c b/src/confd/src/confd/ietf-interfaces.c
index 0cc1bff..c29e54d 100644
--- a/src/confd/src/confd/ietf-interfaces.c
+++ b/src/confd/src/confd/ietf-interfaces.c
@@ -1020,8 +1020,12 @@ static sr_error_t netdag_gen_iface(struct dagger *net,
 
        /* Bring interface back up, if enabled */
        attr = lydx_get_cattr(cif, "enabled");
-       if (!attr || !strcmp(attr, "true"))
+       if (!attr || !strcmp(attr, "true")) {
                fprintf(ip, "link set dev %s up\n", ifname);
+               /* Ensure loopback has operstate UP, not UNKNOWN */
+               if (!strcmp(ifname, "lo"))
+                       fprintf(ip, "link set dev %s state up\n", ifname);
+       }
 
        err = err ? : netdag_gen_sysctl(net, dif);
```

wkz:
I think we can even get away with just this:

```diff
diff --git a/src/confd/src/confd/ietf-interfaces.c b/src/confd/src/confd/ietf-interfaces.c
index 7299ede..142ae3c 100644
--- a/src/confd/src/confd/ietf-interfaces.c
+++ b/src/confd/src/confd/ietf-interfaces.c
@@ -1021,7 +1021,7 @@ static sr_error_t netdag_gen_iface(struct dagger *net,
        /* Bring interface back up, if enabled */
        attr = lydx_get_cattr(cif, "enabled");
        if (!attr || !strcmp(attr, "true"))
-               fprintf(ip, "link set dev %s up\n", ifname);
+               fprintf(ip, "link set dev %s up state up\n", ifname);

        err = err ? : netdag_gen_sysctl(net, dif);
```

troglobit:
Great stuff, lgtm 👍

PR #87, opened 2023-07-11
test/case: Add interface test

ahmkar94:
- Added "iface.py" module for interface related functions(checks)
- Added a test case for interface existence check
- Added test cases for interface parameters "if-index" and "oper-status" check

### DIFF
--- a/test/infamy/iface.py
+++ b/test/infamy/iface.py
@@ -1,0 +1,61 @@
+def extract_iface_param(json_content, param):
+    """Returns (extracted) value for parameter 'param'"""
+    interfaces = json_content.get('interfaces')
+    if not interfaces:
+        return None
+    
+    for interface in interfaces.get('interface'):
+        if param not in interface:
+            continue
+        return interface[param]
+    
+    return None 
+
+def interface_exist(target, iface):
+    """Verify that the target interface exists"""
+    try: 
+        content = target.get_dict(f"/ietf-interfaces:interfaces/ietf-interfaces:interface[name='{iface}']")
+        name = extract_iface_param(content, 'name')
+        return (name != None)
+    except:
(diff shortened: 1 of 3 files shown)


test/infamy/iface.py
```diff
@@ -0,0 +1,54 @@
+def extract_iface_param(row, key):
```
troglobit, line 1: Minor, all class methods should have a docstring right after the `def method` line, for example:

    """Extract foo from bar"""

Also, member names should be descriptive, so it's easier to understand from the method name what they do, which in turn makes it easier to write the docstring.  See more in PEP-8 https://peps.python.org/pep-0008/#documentation-strings

test/infamy/iface.py
```diff
@@ -0,0 +1,54 @@
...
+    interfaces = row.get('interfaces')
+    if interfaces:
+        interface_list = interfaces.get('interface')
+        if interface_list and isinstance(interface_list, list):
+            for interface in interface_list:
+                if key in interface:
+                    return interface[key]
+    return None
```
troglobit, line 9: The staircase effect in this method is a good example of a structure I personally try to avoid.  Exit early and use if-statements to check for error cases, letting the main part of the function continue.

Example pseudo code:

```python
   if Not interfaces:
      return None

   for interface in interfaces.get('interface'):
      if key Not in interface:
         continue
      return interface[key]

   return None
```

troglobit (approved):

test/infamy/iface.py
troglobit: Latest changes look much better.  Great work on this, Ahmed!

troglobit:
Nicely done!  Tests pass (again :), merging!

PR #88, opened 2023-07-12
test/case: upgrade deprecated hash method

ahmkar94:


test/case/ietf_system/add_delete_user.py
```diff
@@ -15,6 +15,12 @@ def generate_restrictred_credential():
 
     return credential
 
+def generate_password_hash(password):
+    command = f"mkpasswd -m SHA-256 {password}"
```
troglobit, line 19: This was a clever idea!  Was expecting a python replacement, but this will work just fine too.  Just to clarify, this is `mkpasswd` from within the container, so the argument to the `-m` option should be one of: `md5`, `sha256`, or `sha512`.  Other versons of the mkpasswd too exist, but this one is the same as we have in Infix (from the BusyBox tool suite).

I suggest adding two arguments to the Python method:
 1. `type="md5"`, argument with default value
 2. `salt`, a random string. If the user provides it, append it to the end of the `command` before running it

test/case/ietf_system/add_delete_user.py
```diff
@@ -23,8 +29,7 @@ def generate_restrictred_credential():
     with test.step("Add new user"):
         username = generate_restrictred_credential()
         password = generate_restrictred_credential()
-        salt = crypt.mksalt(crypt.METHOD_SHA256)
-        hashed_password = crypt.crypt(password, salt)
+        hashed_password = generate_password_hash(password)
```
troglobit, line 32: With my proposal above, the line would then be:

```python
    hashed_password = generate_password_hash(password, type="sha256")
```

troglobit (changes requested):

test/case/ietf_system/add_delete_user.py
ahmkar94: It may look like a clever idea, but I felt I have no other choice :(
I spent too much time trying to find a python function which would do exactly the same thing as "crypt", but at the end none of it worked for me. At least couldn't make it using: bcrypt, passlib, hashlib, argon2 and cryptography.
When it comes to passlib and argon2 at least, I couldn't even get it to the environment properly (install + import weren't enough just enough)... If I knew how to import it properly, maybe that one would work...

troglobit: > [...] I couldn't even get it to the environment properly (install + import weren't enough just enough)... If I knew how to import it properly, maybe that one would work...

Sorry to hear that!  I'm guessing there's a few libraries missing from the python install in the [docker container](https://github.com/kernelkit/infix/blob/main/test/docker/Dockerfile).  For future reference, of if you want to just know how one can update the container: see the [test/docker/README](https://github.com/kernelkit/infix/blob/main/test/docker/README.md)

troglobit: > Thnx for the comments, and if you think it is important to have a python function here, we can spend some more time. Otherwise, I'll just make an update as you proposed in the comments :)

This is a perfectly fine solution, and also verifies the same function (mkpasswd) that we use in Infix.  You definitely don't have to spend any more frustrating hours on it! <3

test/case/ietf_system/add_delete_user.py
```diff
@@ -23,8 +31,7 @@ def generate_restrictred_credential():
     with test.step("Add new user"):
         username = generate_restrictred_credential()
         password = generate_restrictred_credential()
-        salt = crypt.mksalt(crypt.METHOD_SHA256)
-        hashed_password = crypt.crypt(password, salt)
+        hashed_password = generate_password_hash(password, type="sha-256")
```
troglobit, line 34: Sorry, just a final comment.  This works, but only because `mkpasswd` parses `sha` and falls back to `sha256`.  The possible args are: `md5`, `sha256`, and `sha512`.  You can just fix the arg, and put the rest as a function description, within """...""".

troglobit (changes requested):

test/case/ietf_system/add_delete_user.py
ahmkar94: I tried sha256 but the function did not generate hash at all. I think proper syntax needs dash "-" between
image

troglobit: It's the unfortunate difference between the standard Linux (Debian/Ubuntu) mkpasswd and the one shipped in Alpine Linux, which is built around the BusyBox family of tools:
```
$ make test-qeneth-sh
make[1]: Entering directory '/home/jocke/src/infix/buildroot'
/home/jocke/src/infix/test/env -f /home/jocke/src/infix/output/images/infix-x86_64.img -p /home/jocke/src/infix/output/images/infix-x86_64.pkg -q /home/jocke/src/infix/test/virt/dual /bin/sh
Info: Generating topology
Info: Generating node YAML
Info: Generating executables
Info: Launching dut1
Info: Launching dut2
(.infix-test-venv) /home/jocke/src/infix/buildroot # mkpasswd -h
mkpasswd: unrecognized option: h
BusyBox v1.36.0 (2023-05-05 06:41:49 UTC) multi-call binary.

Usage: mkpasswd [-P FD] [-m TYPE] [-S SALT] [PASSWORD] [SALT]

Print crypt(3) hashed PASSWORD

	-P N	Read password from fd N
	-m TYPE	des,md5,sha256/512 (default sha512)
	-S SALT
(.infix-test-venv) /home/jocke/src/infix/buildroot # 
```

test/case/ietf_system/add_delete_user.py
```diff
@@ -2,11 +2,12 @@
 
 import infamy
 import copy
-import crypt
+from passlib.hash import sha256_crypt
```
troglobit, line 5: You need to bump the container version at the top of `test/env` in this PR as well. Otherwise passlib will not be found.

(I'm a bit delayed pushing the new container from PR #89 due to network issues at the office. Sorry!)

troglobit (changes requested):

test/case/ietf_system/add_delete_user.py
ahmkar94: I forgot to rebase after the previous PR.
Pushed again

troglobit: I've just (finally) pushed v0.6 of the container as well, so hopefully things will go much smoother this time :-)

troglobit (approved):

troglobit:
There we go!  Nice job and collab on this one :-)

PR #90, opened 2023-07-23
CLI usability improvments and extensions

### LINKED ISSUE #68: klish `?` only lists first line of yang model's description
Annoyingly, the `klish` (CLI) program only displays the first line of a setting's description (extracted from the YANG model) when the user taps the `?` key.

We should change the the generated help text to append an `...` (ellipsis) to descriptions that contain more text, and then add a `help model container setting` command that can show the complete description.

### LINKED ISSUE #78: klish: add support for 'show interfaces'

troglobit:
This massive PR is a major refresh of the CLI.  It builds on the `cli-rpc-commands` branch, PR #81, and unfortunately has a conflict with another branch, PR #83.  There's also a first version of privsep in PR #71, but think we can just discard that one. Just let me know if there are any conflicts and I'll rebase when the other(s) have been approved/merged.

The majority of the changes are related to the `klish` and `klish-plugin-sysrepo` components.  Both have been bumped to the latest (usable) versions (hashes) from upstream, and both now use a `kkit` branch with fixes and extensions to serve the rest of the changes on this branch.

For klish the changes have been focused on usability.  Standard line editing features have been added, most of which had been disabled upstream when they added UTF8 support, so re-enabling them was not that difficult.

For klish-plugin-sysrepo, the bump to latest git brought in improved online help using `?` and a help text generator was added to extract the YANG model description for leaf nodes.  As an experimental feature, an `srp_prompt` symbol was also added to replace the JunOS style `srp_prompt_edit_path` symbol in favor of an inline path-style prompt resembling a regular shell.

With the fixes and feature extensions to klish in place, line editing and movement in the CLI had been greatly improved, a few earlier decisions regarding behavior and key mapping started becoming annoying.  So, the most controversial change in this patch set includes:

 - The mapping of Ctrl-C has been restored to abort editing of the current line
 - To exit/abort from a sub-context, Ctrl-D is now the key, as long as the line is empty otherwise Ctrl-D deletes the character to the right of the cursor

This is more inline with how a regular shell works, except for the overloaded Ctrl-D mapping.

Additionally, it addresses the following issues:

 - Fix #78, initial support by calling `ip link show` with optional arguments
 - Fix #68, support for a new `help` command.  In configure context shows Yang description
 - The auto-factory-default.cfg file is now set to read-only (0444) to prevent accidental deletion, e.g., from a future copy command extension
 - Default editor is now Mg, with a usability patch included for the short-help text in status buffer
 - CLI documentation is now extracted from the doc/cli/ directory of the repo and made available from admin-exec using the new `help [introduction | keybindings | tutorial]` command

There is more, but in comparison very minor changes.


wkz:
Great work!

@troglobit: Am I correct to assume that the privsep problems are not fully solved by this PR? I.e. forked child now runs with the correct UID/GID, but from sysrepo's POV it has root access to datastores?

troglobit:
> Great work!

Thanks, went a bit nuts but the overall result turned out OK 😊 

> @troglobit: Am I correct to assume that the privsep problems are not fully solved by this PR? I.e. forked child now runs with the correct UID/GID, but from sysrepo's POV it has root access to datastores?

That is correct. I've commented out that part and added a ticket to the board for follow-up later.  At least the forked-off CLI daemon tracks the UID so the prompt is correct. Current behavior is OK but could be improved upon.  It should be possible to go all the way, and drop privileges completely, I estimate another couple of days for that.

PR #91, opened 2023-07-31
statd: add interface type to operational data

rical:
An iana or infix interface type is derived from "ip link" data. Such as loopback -> iana-if-type:softwareLoopback.

This patch adds support for the following types:
loopback, ethernetCsmacd, veth, l2vlan and bridge.

Example data:
```
{
  "ietf-interfaces:interfaces": {
    "interface": [
      {
        "name": "lo",
        "type": "iana-if-type:softwareLoopback",
        "oper-status": "unknown",
        "if-index": 1
      },
      {
        "name": "e0",
        "type": "iana-if-type:ethernetCsmacd",
        "oper-status": "up",
        "if-index": 2
      },
      {
        "name": "e0.100",
        "type": "iana-if-type:l2vlan",
        "oper-status": "down",
        "if-index": 3
      },
      {
        "name": "br0",
        "type": "iana-if-type:bridge",
        "oper-status": "down",
        "if-index": 4
      },
      {
        "name": "veth1",
        "type": "infix-if-type:veth",
        "oper-status": "down",
        "if-index": 5
      }
    ]
  }
}
```

### DIFF
--- a/src/statd/statd.c
+++ b/src/statd/statd.c
@@ -247,10 +247,67 @@ err_out:
 	return SR_ERR_SYS;
 }
 
+static const char *get_yang_link_type(char *xpath, json_t *iface)
+{
+	json_t *j_type;
+	const char *type;
+
+	j_type = json_object_get(iface, "link_type");
+	if (!json_is_string(j_type)) {
+		ERROR("Expected a JSON string for 'link_type'");
+		/* This will throw a YANG / ly error */
+		return "";
+	}
+
+	type = json_string_value(j_type);
+
+	if (strcmp(type, "ether") == 0) {
+		const char *kind;
+		json_t *j_val;
+
+		j_val = json_object_get(iface, "linkinfo");
+		if (!j_val)
+			return "iana-if-type:ethernetCsmacd";
+
+		j_val = json_object_get(j_val, "info_kind");
+		if (!j_val)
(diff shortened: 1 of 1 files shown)


troglobit:
@wkz what do you think, should this perhaps be derived from the interface type in the running datastore instead of duplicated in operational?

rical:
> @wkz what do you think, should this perhaps be derived from the interface type in the running datastore instead of duplicated in operational?

Isn't the running datastore based on configuration data? I thought the idea with operational data was to get it directly from the source i.e. the running kernel.

troglobit:
> 

Yes mostly, and that's the crux of it.  The interface type is derived from a) physical interfaces that exist at first boot (factory-config), like ethernetCsmacd for all ethN/eN interfaces, but also loopback.  Other types are inferred from the name when creating them, e.g. veth pairs.  See `ifchange_cand_infer_veth()`.

I'm just concerned we start bookkeeping this core knowledge in more than one place, so curious what @wkz thinks about it.

troglobit:
Good points all around, and I agree. I'm just worried about duplication and keeping it all in sync. I guess we should put things like this in a checklist or auto test to ensure we don't drift apart in the various backends. 

Also, see https://github.com/addiva-elektronik/alder/issues/283

PR #92, opened 2023-08-01
statd: add support for additional operstates

rical:
Add support for yang link state "lower-layer-down" and "not-present".

### DIFF
--- a/src/statd/statd.c
+++ b/src/statd/statd.c
@@ -60,12 +60,6 @@ static void set_sock_rcvbuf(int sd, int size)
        DEBUG("Socket receive buffer size increased to: %d bytes\n", size);
 }
 
-static void str_to_lowercase(char *str)
-{
-	for (int i = 0; str[i]; i++)
-		str[i] = tolower((unsigned char)str[i]);
-}
-
 static int nl_sock_init(void)
 {
 	struct sockaddr_nl addr;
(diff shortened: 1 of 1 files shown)


src/statd/statd.c
```diff
@@ -145,11 +139,25 @@ static json_t *json_get_ip_link(char *ifname)
...
+	if (strcmp(operstate, "DOWN") == 0)
+		return "down";
+	if (strcmp(operstate, "UP") == 0)
+		return "up";
+	if (strcmp(operstate, "LOWERLAYERDOWN") == 0)
+		return "lower-layer-down";
+	if (strcmp(operstate, "NOTPRESENT") == 0)
+		return "not-present";
+
+	return "unknown";
```
troglobit, line 153: This fixes the issue I saw with statd crashing on unknown strings, but is a regression to the `str_to_lowercase()` in that it maps `dormant` and `testing` to `unknown`.

troglobit (changes requested):
Overall better solution to any of what we have right now.  I did a few more fixes in #82 that I'd like to get in, but I could also rework that PR.

Just a thought, instead of multiple `strcmp()`, personally I'd go for more of a table of strings:
```C
struct {
  char *kern;
  char *yang;
} map[] = {
  { "foo", "bar" },
};

for (size_t i = 0; sizeof(map) / sizeof(map[0]); i++) {
  if (strcmp(operstate, map[i]))
    continue;
  return map[i];
}

return "unknown";

troglobit (approved):
Nice work! :+1:

troglobit:
Btw, I'll be back from vacation in a couple of week, but iirc @wkz will be back already next week, Monday.

PR #93, opened 2023-08-07
statd: add tx/rx counters to iface operational data

rical:
Add tx and rx bytes as YANG out-octets and in-octets.


troglobit (approved):
lgtm

PR #105, opened 2023-08-16
Fine tune readme

troglobit:
result

Some squeeze and tweak with a bit of `git mv` and `git rm` to clean up unnecessary files from the root directory.


wkz:
Shiny! :sparkles:

PR #112, opened 2023-08-21
Misc YANG fixes

troglobit:
- Fix lost deviation and timezone enumerations
- Merge internal YANG revisions inside release cycle
- Drop old versions of native YANG models
- Lock down and restrict user shells
- Limit if:type to supported native types only


troglobit:
Rebased against `main` and force-pushed to get and update the `routing_basic.py` test case, which was failing in previous runs.

wkz (changes requested):
As discussed: Drop must expr on iftype, instead set base type to infix-iftype.

Otherwise brilliant as always

troglobit:
Dropped the first commit, introducing the must expression, then adjusted the last if:type limiting commit, dropping the extension to the same must expression.

PR #114, opened 2023-08-22
Fix #107: auto-sync forks every day 2:42 UTC or on dispatch

### LINKED ISSUE #107: New sync-fork action
Should be triggered by cronjob and require a condition variable to be set in the GitHub settings of the fork.

troglobit:
Highly experimental.


troglobit:
As discussed, added condition variable.

PR #117, opened 2023-08-22
Fixing broken urls in document docs

jovatn:
Fixing some broken links.
Regarding the url to the CLI guide in container.md, changing it to cli/README.md
(rather than cli/introduction.md) is perhaps better.  Or to "cli/"!

I have a problem to display https://github.com/jovatn/infix-jovatn/tree/main/doc/cli/README.md
but with https://github.com/jovatn/infix-jovatn/tree/main/doc/cli it works good


troglobit (approved):
Brilliant, thanks!  Will merge as soon as the regression tests have passed.

Issue #118, opened 2023-08-22
package/skeleton-init-finit: enable fewer services by default


troglobit:
The services enabled by default in Infix should not include those which we plan on adding YANG support for (short-term). I propose the following services be enabled by default, for now:

| Service | Comment |
|---------|---------|
|`getty` ||
|`dnsmasq`  | Local resolver, set up in ietf-system |
|`openssh` ||
|`ospfd`  | + `zebra/etc.`, only if Quagga/Frr is enabled |
|`sysklogd` ||
|`watchdog`||

These are enabled today, so I propose disabling the following:

| Service | Comment |
|---------|---------|
|`avahi` | mDNS/SD should be in discover/mdns yang model |
|`lldpd` | LLDP should be in std IEEE yang model |
|`snmpd` | SNMP should be in std IETF yang model |
|`nginx` | Later when we add RESTCONF support |
|`sshdpd` | SSDP should in discover/ssdpd yang model |
|`gdbserver` | GDB remote server[^1], should possibly be in diagnose/gdb-server[^2] |

[^1]: Actually broken rn after qemu/run-menuconfig refactor
[^2]: Another friend in diagnose yang is the inet echo server for Achilles testing

Issue #116, opened 2023-08-22
confd: break out srx fns to separate library


troglobit:
Other plugins and daemons (out of tree) might want to enjoy the neat helpers created in confd.

PR #119, opened 2023-08-22
Configurable services

troglobit:
1. Disable all services we want to have configurable (#118)
2. Refactor to break out a `libsrx` from `confd` (#116)
3. Flatten `confd` file hierarchy (#115)
4. YANG support for enabling/disabling mDNS/SD and SSDP
5. Enable mDNS/SD and SSDP by default in `factory-config`
6. Enable LLDP by default in `factory-config`

There's also a few other minor fixes and additions I found while cleaning up my working directory after the summer, should not be too much of a surprise, I hope.

As discussed offline, any native SNMP support in Infix is postponed for a future date.


troglobit:
As discussed in our AFK code review:

 - drop `discover/` level from `mdns` and `ssdp` in the yang model, placing them on the root level along with `lldp` and `snmp`
 - add LLDP service support (heavily redacted) config using ieee802-dot1ab-lldp std yang model
 - enable LLDP in Infix factory-config

Other changes compared to the original PR:

 - RFC: add CLI `end` alias to `up`, because that's what I've got in my bones
 - Rename `sysrepo-bootstrap.sh` -> `confd-bootstrap.sh`, it lives in `src/confd/` ...

troglobit:
Fix copy-paste error in factory-config.  Locally checked and rechecked this time.

troglobit:
Converted to draft, because we need to discuss the NETCONF wire format for enabled.  Right now we have `infix-services:enabled` for mdns and ssdp, but `infix-lldp:enabled` for lldp.  I suggest we:

  a) either split infix-services.yang in two and standardize on `infix-<SVC>:enabled`, or
  b) add a `infix-service:enabled` typedef that's used everywhere.

test/infamy/__init__.py
```diff
@@ -3,6 +3,7 @@
 from .env import Env
 from .netns import IsolatedMacVlan
 from .sniffer import Sniffer
+from .ssdp import SsdpClient
```
wkz, line 6: This seems specific enough that we might leave it out of the default case.

I.e. the tests that need it can `from infamy.ssdp import SsdpClient` explicitly.

test/case/infix_services/services_basic.py
```diff
@@ -12,25 +17,33 @@ def verify(enabled, sec):
...
+        target.put_config_dict("ieee802-dot1ab-lldp", {
+            "lldp": {
+                "enabled": enabled
+            }
+        })
 
-            #running = target.get_config_dict("/infix-services:mdns")
-            #assert running["mdns"]["enabled"] == enabled
+        #running = target.get_config_dict("/ieee802-dot1ab-lldp:lldp")
+        #assert running["lldp"]["enabled"] == enabled
```
wkz, line 44: Not needed I guess?

test/case/infix_services/services_basic.py
```diff
@@ -72,24 +85,33 @@ def verify(enabled, sec):
                 "enabled": False
             }
         })
+        target.put_config_dict("ieee802-dot1ab-lldp", {
+            "lldp": {
+                "enabled": False
+            }
+        })
         cfg = target.get_config_dict("/infix-services:ssdp")
         assert not cfg["ssdp"]["enabled"]
```
wkz: For symmetry, should we add the same checks for mDNS and LLDP?

src/confd/yang/infix-services@2023-08-22.yang
```diff
@@ -0,0 +1,36 @@
...
+  revision 2023-08-22 {
+    description "Initial revision, add SSDP and mDNS/SD enable/disable only.";
+    reference "internal";
+  }
+
+  /*
+   * Data nodes
+   */
+
+  container mdns {
```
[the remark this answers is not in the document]
troglobit: Hmm, didn't consider that, but it feels like any such config should be from within a `container dns` or `resolver` context, right?

test/infamy/__init__.py
troglobit: Good point, didn't really know what `__init___.py` was for, will drop in a follow-up commit.

test/case/infix_services/services_basic.py
troglobit: Maybe, or drop and in the future replace with an operational check?

test/infamy/__init__.py
wkz: You mean you'll add it to this PR, or to a later one?

test/case/infix_services/services_basic.py
wkz: Even better.

test/infamy/__init__.py
troglobit: To this PR, sorry for betting so slow.

wkz: Yeah, the famously unproductive @troglobit! :wink: 

Take all the time you need, I just didn't want to leave this PR in limbo :+1:

PR #121, opened 2023-08-24
Aarch64 devboard fixes

wkz:
With this series applied, current Infix images can be run on:

- SparX-5i PCB135, using the "Traditional FIT" image. The regular image does not work as we are still relying on the old U-Boot shipped by Microchip, which does not have the blkmap support we need.
- CN9130-CRB, using the regular image. U-Boot has been bumped to the latest version (2023.07.02)

The last change in this series introduces the concept of "internal" ports (DSA ports) which should not be managed by confd. For now, we simply make sure that internal interfaces don't end up in the factory-config - long-term we might want to add further protections to confd to forbid it from configuring these interfaces.


troglobit (approved):
lgtm

PR #122, opened 2023-08-28
Proposal for how to document netoworking use-cases/capabilities

troglobit:
Here's my idea for how to document use-cases, and/or networking capabilities of Infix. 
Ping @jovatn + @wkz 

Preview here: https://github.com/kernelkit/infix/blob/interface-lego/doc/networking.md


jovatn:
Note that wkz has updated the look of the lego pieces a bit. See svg examples in #137

troglobit:
> Note that wkz has updated the look of the lego pieces a bit. See svg examples in #137

Yup, updating to that now, thanks!

PR #124, opened 2023-08-29
License fixes for local packages

troglobit:
Minor cleanup of licenses and packaging for `statd` + `confd`.

> Note: branch based on `configurable-services`, PR #119


troglobit:
For some reason, one of the jobs fail.  I'm restarting it for the second time now ... 

... seems to be LLDP not showing up when being enabled.  I've considered improving that by adding socat to the Docker container, but we're a bit short for time, for the release, so I put it off for now.  What do you think, this particular branch does not affect that part of the code, but builds on `configurable-services` which is a blocker for the release?

Issue #127, opened 2023-08-30
Drop x86_64-minimal defconfig


troglobit:
Use std defconfig instead for regression tests.

https://github.com/kernelkit/infix/blob/8fc709676fed3ca04727c25f3704196ad564d1dc/.github/workflows/regression.yml#L46-L48

PR #134, opened 2023-09-12
test: add top level unit test suite

rical:
This currently includes cli-pretty tests. It could be executed separately to clarify test results in github actions. This commit includes it in the main test execution. The most important thing is that it's executed.

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -19,6 +19,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          submodules: 'true'
       - name: Set Build Variables
         id: vars
         run: |
--- a/board/x86_64/board.mk
+++ b/board/x86_64/board.mk
@@ -1,5 +1,5 @@
 test-dir := $(BR2_EXTERNAL_INFIX_PATH)/test
-INFIX_TESTS ?= $(test-dir)/case/all.yaml
+INFIX_TESTS ?= $(test-dir)/case/all-unit.yaml $(test-dir)/case/all.yaml
 
 test-env = $(test-dir)/env \
 	-f $(BINARIES_DIR)/infix-x86_64.img \
--- a/test/case/all-unit.yaml
+++ b/test/case/all-unit.yaml
@@ -1,0 +1,5 @@
+---
+# Tests in this suite can be run on localhost without a target environment
+
+- name:  cli-pretty
+  suite: cli_pretty/all.yaml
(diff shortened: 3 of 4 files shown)


troglobit:
It fails because of "file not found" for missing `test/case/pretty_cli/run.sh` but that seems to be there, so my guess is that this fails because Reggie Regression (github job) still uses the minimal defconfig.

If you include a fix to #127 in your PR I suspect this will fix the problem.

rical:
> It fails because of "file not found" for missing `test/case/pretty_cli/run.sh` but that seems to be there, so my guess is that this fails because Reggie Regression (github job) still uses the minimal defconfig.
> 
> If you include a fix to #127 in your PR I suspect this will fix the problem.

Thanks for investigating. I pushed a new version that **won't** fix the problem but includes some new 9pm commits for more graceful error checking of missing files. Lets see if we at least can make the error look better ;)

troglobit:
This still fails with:

```
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/9pm_e_2h98lp/9pm_iusksjh5_db'
make[1]: *** [/home/runner/work/infix/infix/board/x86_64/board.mk:15: test-qeneth] Error 1
```

Do we have any idea why?

rical:
> This still fails with:
> 
> ```
> FileNotFoundError: [Errno 2] No such file or directory: '/tmp/9pm_e_2h98lp/9pm_iusksjh5_db'
> make[1]: *** [/home/runner/work/infix/infix/board/x86_64/board.mk:15: test-qeneth] Error 1
> ```
> 
> Do we have any idea why?

Apparently there's no bash interpreter? I changed the script `run.sh` to call `#!/usr/bin/env bash` which revealed the real problem: `env: can't execute 'bash': No such file or directory`. I tried adding it in `regression.yml` as `shell: bash`, but no luck.

troglobit:
> > This still fails with:
> > ```
> > FileNotFoundError: [Errno 2] No such file or directory: '/tmp/9pm_e_2h98lp/9pm_iusksjh5_db'
> > make[1]: *** [/home/runner/work/infix/infix/board/x86_64/board.mk:15: test-qeneth] Error 1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do we have any idea why?
> 
> Apparently there's no bash interpreter? I changed the script `run.sh` to call `#!/usr/bin/env bash` which revealed the real problem: `env: can't execute 'bash': No such file or directory`. I tried adding it in `regression.yml` as `shell: bash`, but no luck.

Very odd, even the minimal defconfig has bash ... or wait, is this for executing inside the Alpine Docker?  If so, I don't think we have bash there, only 'ash' the default BusyBox shell included in the basic Alpine install.

troglobit:
There we go, all PASS! :-)

troglobit (approved):

PR #135, opened 2023-09-13
add ip origin tracking

rical:
This PR adds ip proto / origin info.

```
root@infix-00-00-00:/> show interfaces 
INTERFACE        PROTOCOL          STATE          DATA                          
e0               ethernet          UP             02:00:00:00:00:00             
                 ipv4                             192.168.1.1/24 (static)
                 ipv4                             198.10.1.162/24 (dhcp)
lo               ethernet          UP             00:00:00:00:00:00             
                 ipv4                             127.0.0.1/8 (static)

```

It does this by setting ip/kernel `proto` and converting it to YANG `origin`, which is exposed to the user.

**NOTE**
* The udhcpc default.script is modified in this PR, I have tested it manually and it works better than before (see commit msg). However, such modifications are error-prone and we need to beef up testing in this area.

### DIFF
--- a/board/common/rootfs/usr/share/udhcpc/default.script
+++ b/board/common/rootfs/usr/share/udhcpc/default.script
@@ -1,0 +1,122 @@
+#!/bin/sh
+
+# udhcpc script edited by Tim Riker <Tim@Rikers.org>
+
+[ -z "$1" ] && echo "Error: should be called from udhcpc" && exit 1
+
+ACTION="$1"
+RESOLV_CONF="/etc/resolv.conf"
+[ -e $RESOLV_CONF ] || touch $RESOLV_CONF
+[ -n "$broadcast" ] && BROADCAST="broadcast $broadcast"
+[ -n "$subnet" ] && NETMASK="netmask $subnet"
+# Handle stateful DHCPv6 like DHCPv4
+[ -n "$ipv6" ] && ip="$ipv6/128"
+
+if [ -z "${IF_WAIT_DELAY}" ]; then
+	IF_WAIT_DELAY=10
+fi
+
+wait_for_ipv6_default_route() {
+	printf "Waiting for IPv6 default route to appear"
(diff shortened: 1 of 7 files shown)


board/common/rootfs/usr/share/udhcpc/default.script
```diff
@@ -0,0 +1,134 @@
...
+		if [ -x /usr/sbin/avahi-autoipd ]; then
+			/usr/sbin/avahi-autoipd -c $interface || /usr/sbin/avahi-autoipd -wD $interface --no-chroot
+		fi
+		;;
+
+	renew|bound)
+		if [ -x /usr/sbin/avahi-autoipd ]; then
+			/usr/sbin/avahi-autoipd -c $interface && /usr/sbin/avahi-autoipd -k $interface
+		fi
+		/bin/ip addr add dev $interface $ip/$NETMASK $BROADCAST proto 5
```
troglobit, line 72: For `proto 5`, and `proto 4` below is going to become pretty nasty.  I thought it would be possible to use the names from rt_addrprotos instead?

rical: Strangely the upstream code doesn't allow ASCII when setting proto.
The manual states `ADDRPROTO := [ NAME | NUMBER ]` but adding a string throws `Error: argument "static" is wrong: "proto" value is invalid`. Checking the code it does:
```                                                                                                                                                     
172 +  if (get_u8(&proto, *argv, 0))                                                                                                                                        
173 +      invarg("\"proto\" value is invalid\n", *argv);
```

Maybe this is fixed in later upstream patches. The number does however resolve to ASCII during printing (`ip addr show`). I think we can fix this in a later patchset?

troglobit: Aha, yeah must be a missing patch upstream.  Right, let's fix that later!

PR #139, opened 2023-09-18
confd: fix IPv6 autoconf support

### LINKED ISSUE #136: IPv6 address autoconf 'create-global-addresses true' does not bite
The yang/cli setting for autoconfiguring global addresses does not seem to bite.
I.e., the following will not give me a global address even when receiving router advertisements on eth0
```
root@infix-00-00-00:/config/> set interfaces interface eth0 ipv6 autoconf create-global-addresses true
```
A look in proc shows the following.
```
root@infix-00-00-00:~$ cat /proc/sys/net/ipv6/conf/eth0/autoconf 
0
```

Flipping it to "1"  and taking eth0 down/up will give me the global address
```
root@infix-00-00-00:~$ echo 1 > /proc/sys/net/ipv6/conf/eth0/autoconf 
root@infix-00-00-00:~$ ip link set down dev eth0
root@infix-00-00-00:~$ ip link set up dev eth0
root@infix-00-00-00:~$ ip a show dev eth0
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether 02:00:00:00:00:00 brd ff:ff:ff:ff:ff:ff
    inet6 2001:db8:0:1:0:ff:fe00:0/64 scope global dynamic mngtmpaddr               <=======
       valid_lft 86399sec preferred_lft 14399sec
    inet6 fe80::ff:fe00:0/64 scope link 
       valid_lft forever preferred_lft forever
root@infix-00-00-00:~$ 
```

In this case I had a router advertising prefix 2001:db8:0:1::/64

```
linux-router # cat /etc/radvd.conf 
interface tap-infix
       {
               AdvSendAdvert on;
               prefix 2001:db8:0:1::/64
               {
                       AdvOnLink on;
                       AdvAutonomous on;
               };
       };

linux-router #
```

troglobit:
- Enable privacy feature, allows true random address instead of EUI64
 - Allow global addresses to use router prefix

Fixes #136

### DIFF
--- a/src/confd/confd-bootstrap.sh
+++ b/src/confd/confd-bootstrap.sh
@@ -56,6 +56,7 @@ sysrepoctl -s $SEARCH							\
 	   -i ietf-interfaces@2018-02-20.yang   -g wheel -p 0660	\
 		-e if-mib						\
 	   -i ietf-ip@2018-02-22.yang		-g wheel -p 0660	\
+		-e ipv6-privacy-autoconf				\
 	   -i ietf-network-instance@2019-01-21.yang -g wheel -p 0660	\
 	   -i ietf-netconf-monitoring@2010-10-04.yang -g wheel -p 0660	\
 	   -i ietf-netconf-nmda@2019-01-07.yang -g wheel -p 0660	\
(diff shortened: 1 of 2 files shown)


wkz (approved):
:tada:

PR #141, opened 2023-09-20
Add support for custom phys-address

troglobit:


troglobit:
So obviously I forgot to add the new test to the suite, fixed now.

src/confd/src/ietf-interfaces.c
```diff
@@ -416,6 +416,36 @@ static int netdag_set_conf_addrs(FILE *ip, const char *ifname,
...
+	const char *ifname = lydx_get_cattr(dif, "name");
+	const char *mac = NULL;
+	struct lyd_node *node;
+	char buf[32];
+
+	node = lydx_get_child(dif, "phys-address");
+	if (lydx_get_op(node) == LYDX_OP_DELETE) {
+		FILE *fp;
+
+		fp = popenf("r", "ethtool -P %s |awk '{print $3}'", ifname);
```
wkz, line 430: This does not seem to be universally supported (e.g. on `veth` and `bridge` interfaces):

```
$ ethtool -P virbr0 | awk '{ print($3); }'
not
```
So what do we do when a custom MAC config is deleted from a virtual interface?

troglobit: Good point, this might be one of those instances where we need to trigger `netdag_must_del()`.   I'll look into it and get back to you!

Also, I just found that `ip link show` shows the inverse, i.e., when there is an address to restore to it lists: `permaddr aa:bb:cc:dd:ee:ff`, so we don't need `ethtool` at least.

troglobit: Fixed by extending `netdag_must_del()` with another condition for all affected interfaces.

troglobit:
Fixes:

 1. replaced `ethtool .. | awk` with `ip link ... | jq`
 2. recreate interfaces that don't support permaddr attribute
 3. some minor log cleanup from previous PRs

wkz (approved):
:shipit:!

troglobit:
> :shipit:!

Thanks :clinking_glasses: will merge to main when the CI job has completed.

PR #144, opened 2023-09-25
statd: add support for ietf-interfaces/interface/admin-status

troglobit:
Xlate IFF_UP flag to admin-status: up/down.  We do not support any 'testing' mode (yet) in Infix.

### DIFF
--- a/src/statd/statd.c
+++ b/src/statd/statd.c
@@ -391,6 +391,26 @@ static int ly_add_ip_link_br(const struct ly_ctx *ctx, struct lyd_node **parent,
 	return SR_ERR_OK;
 }
 
+static int iface_has_flag(json_t *j_iface, const char *flag)
+{
+	json_t *j_flags, *j_flag;
+	size_t i;
+
+	j_flags = json_object_get(j_iface, "flags");
+	if (!j_flags)
+		return 0;
+
+	if (!json_is_array(j_flags))
+		return 0;
+
+	json_array_foreach(j_flags, i, j_flag) {
+		if (strcmp(json_string_value(j_flag), flag) == 0)
+			return 1;
+	}
+
+	return 0;
+}
+
 static int ly_add_ip_link_data(const struct ly_ctx *ctx, struct lyd_node **parent,
 			       char *xpath, json_t *j_iface)
 {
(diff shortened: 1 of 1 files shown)


src/statd/statd.c
```diff
@@ -391,6 +391,31 @@ static int ly_add_ip_link_br(const struct ly_ctx *ctx, struct lyd_node **parent,
 	return SR_ERR_OK;
 }
 
+static int iface_has_flag(json_t *j_iface, const char *ifflag)
+{
+	json_t *flags, *flag;
+	int found = 0;
```
rical, line 397: I don't see the need for this flag? Why not just return 1 when you find the flag?
Something like:
```
if (strcmp(json_string_value(j_flag), flag) == 0)
    return 1
```

src/statd/statd.c
```diff
@@ -391,6 +391,31 @@ static int ly_add_ip_link_br(const struct ly_ctx *ctx, struct lyd_node **parent,
...
+
+	flags = json_object_get(j_iface, "flags");
+	if (!flags)
+		return 0;
+
+	if (!json_is_array(flags))
+		return 0;
+
+	json_array_foreach(flags, i, flag) {
+		ERROR("Comparing flag %s vs ifflag %s", json_string_value(flag), ifflag);
```
rical, line 408: Accidental debug left in the code?

src/statd/statd.c
```diff
@@ -391,6 +391,31 @@ static int ly_add_ip_link_br(const struct ly_ctx *ctx, struct lyd_node **parent,
 	return SR_ERR_OK;
 }
 
+static int iface_has_flag(json_t *j_iface, const char *ifflag)
```
rical, line 394: If you want to align the variable naming to the rest of the statd code, this should be `const char *flag` and the json flags below should be `j_flags`and `j_flag`. No big deal.

rical (commented):
Minor style comments

src/statd/statd.c
troglobit: Ah, yes. Will fix.

src/statd/statd.c
troglobit: OK, will fix!

rical (approved):

PR #146, opened 2023-09-25
statd: split code into multiple files (cosmetic)

rical:
Mainly in preparation for upcoming patches for ethtool.

There is no functional change.


troglobit (approved):
Looks great!  Pending tests before merging ...

PR #147, opened 2023-09-27
draft: add basic ethtool statistics support

rical:
## Statd 
This PR adds the framework for reading ethtool statistics and inserting it into the operational datastore.

The ethtool data we rely on is "group data" such as "eth-mac" or "rmon".

This data can be displayed using:
`ethtool --json -S e0 --all-groups`

The "group data" is still missing for most common drivers, so testing this will require a firmware which has support for it.

## CLI
This PR also adds support for printing all Ethernet frame statistics which exists in the operational datastore.
```
> show interfaces name e0
....                                                                                                                        
                                                                                                                                                                             
eth-in-frames       : 222                                                                                                                                                    
eth-out-frames      : 1337
```

## TODO
Test on HW.


src/confd/yang/ieee802-ethernet-interface.yang
troglobit, line 1: Everything looks good to me, except for the YANG file.  We've said we should version all our YANG models, even if they are not versioned upstream.  So please add `@2019-06-21` to it, according to its revision log, and update `confd-bootstrap.sh` accordingly.

troglobit (changes requested):

troglobit:
Checks fail still, with the same unstable tests (MTU/operational), did you rebase on the latest `main` before force-pushing?

rical:
> Checks fail still, with the same unstable tests (MTU/operational), did you rebase on the latest `main` before force-pushing?

Yes I did, strange.

troglobit:
Odd indeed, wondering if there's something similar to before, that statd bugs out because of some YANG model limitation/deviation we've added elsewhere.  

I can have a look at your branch a bit later, in the middle of a bug of my own at the moment.

troglobit:
We've now merged #161, with the necessary kernel upgrade and driver fixes for RMON counter.

troglobit:
> > We've now merged #161, with the necessary kernel upgrade and driver fixes for RMON counter.
> Great! I rebased and built `aarch64_defconfig` from git clean. But I don't get any counters in `ethtool --json -S eth1 --all-groups`. Which is the origin of the info in this PR. However I do see a lot of counters in `ethtool -S eth1`. Investigation ongoing.

Hmm, hopefully @wkz can weigh in here.

troglobit:
After fixing #127 on this branch, tests PASS https://github.com/rical/infix/actions/runs/6548703982.  There still seems to be some merge conflict so I'll have a look offline and merge it to main manually.

Nice work on this, and thank you for your patience!

troglobit:
Rebased and merged to main.

PR #148, opened 2023-09-27
statd: handle IP origin for IPv6 RA

rical:
Handle IPv6 Router Advertisement (RA) based addresses.

If the last 64 bits of a link-local address is generated using the local MAC address of the interface, i.e. a EUI-64 based address. Then the origin is set to "link-layer".

If the addrgenmode is set to "random" instead of "eui64", then the device will generate the interface identifier portion of the address (the last 64 bits) randomly rather than deriving it from the MAC address using the EUI-64 process. This will result in a "random" origin.

### DIFF
--- a/src/statd/iface-ip-addr.c
+++ b/src/statd/iface-ip-addr.c
@@ -20,7 +20,7 @@ static json_t *json_get_ip_addr(const char *ifname)
 	return json_get_output(cmd);
 }
 
-static const char *get_yang_origin(const char *protocol)
+static const char *get_yang_origin(const char *proto, json_t *j_addr)
 {
 	size_t i;
 	struct {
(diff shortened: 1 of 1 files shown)


troglobit (approved):
lgtm :+1: 

**Comment:** we'll probably add support for "stable-privacy" later (24Q3), it's just like random, only resulting in the same address every boot.

jovatn:
IPv6 global addresses shows origin "other" rather than "link-layer" or "random". It should mimic the behavior of the fe80-address, as both use the same address suffix.

```
admin@infix-00-00-00:/> show interfaces 
INTERFACE       PROTOCOL   STATE       DATA                                     
eth0            ethernet   UP          02:00:00:00:00:00                        
                ipv6                   2001:db8:0:1:0:ff:fe00:0/64 (other)
                ipv6                   fe80::ff:fe00:0/64 (link-layer)
lo              ethernet   UP          00:00:00:00:00:00                        
                ipv4                   127.0.0.1/8 (static)
                ipv6                   ::1/128 (static)
admin@infix-00-00-00:/> configure
admin@infix-00-00-00:/config/> set interfaces interface eth0 ipv6 autoconf create-temporary-addresses true 
admin@infix-00-00-00:/config/> leave
admin@infix-00-00-00:/> show interfaces 
INTERFACE       PROTOCOL   STATE       DATA                                     
eth0            ethernet   UP          02:00:00:00:00:00                        
                ipv6                   2001:db8:0:1:ea05:e93e:ad9c:56df/64 (other)
                ipv6                   fe80::f9:2297:2ac:6b65/64 (random)
lo              ethernet   UP          00:00:00:00:00:00                        
                ipv4                   127.0.0.1/8 (static)
                ipv6                   ::1/128 (static)
admin@infix-00-00-00:/> 
```

jovatn:
Apologize for my comment above. Must have been tested on older image.

troglobit:
@rical I've just merged #152, so you can go ahead and rebase on the fixed `main` branch now :+1:

PR #149, opened 2023-10-02
Add deviations for MTU configuration (IPv4/IPv6) in ietf-ip.yang

jovatn:
MTU configuration/status not yet implemented in Infix, thus deviations added.


troglobit (approved):
I'll approve this and let it be part of the v23.09 release since we agreed on it last week. Thanks!

PR #150, opened 2023-10-02
Use GitHub result file in regression test

rical:
Use the new fancy result-gh.md file which as github emojis.
Also,  split result publishing to its own run command to simplify the YAML file.

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -61,9 +61,11 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           path: output/${{ steps.vars.outputs.tgz }}
+
       - name: Regression Test
-        run: |
-          rc=true
-          make test-qeneth || rc=false
-          cat test/.log/last/result.md >> $GITHUB_STEP_SUMMARY
-          eval $rc
+        run: make test-qeneth
+
+      - name: Publish Test Result
+        # Ensure this runs even if Regression Test fails
+        if: always()
+        run: cat test/.log/last/result-gh.md >> $GITHUB_STEP_SUMMARY


.github/workflows/regression.yml
```diff
@@ -61,9 +61,9 @@ jobs:
...
       - name: Regression Test
-        run: |
-          rc=true
-          make test-qeneth || rc=false
-          cat test/.log/last/result.md >> $GITHUB_STEP_SUMMARY
-          eval $rc
+        run: make test-qeneth
+
+      - name: Publish Test Result
+        run: cat test/.log/last/result-gh.md >> $GITHUB_STEP_SUMMARY
```
troglobit, line 69: Neat with more fancy log file! :rocket: 

However, all GitHub actions run with a sort of `set -e`, so with your refactor we'll never get the test results published if any test fails.

troglobit (changes requested):
See review comment inline.

.github/workflows/regression.yml
rical: > Neat with more fancy log file! 🚀
> 
> However, all GitHub actions run with a sort of `set -e`, so with your refactor we'll never get the test results published if any test fails.

Great point. Didn't know that. I added the magical GitHub syntax `if: always()` which ensures the test results are published even if previous jobs fails.

Here's an example from the test framework:
https://github.com/rical/9pm/actions/runs/6377739586

troglobit (approved):

.github/workflows/regression.yml
troglobit: Nice!  Now I can fix a few of my other actions as well, thanks :smiley::+1:

troglobit:
Merging despite regression test failure.  This problem is common to the main branch and tracked in #151.

Issue #151, opened 2023-10-02
Operational status broken in v23.09


troglobit:
With the merge of #149 to `main` we broke `statd`, because it turns out we have implemented read-back of MTU in operational data, although configure support is not yet available.  So the change to disable the MTU YANG nodes cause the following error:

```
Oct  2 12:19:46 infix-00-00-00 statd: lydx_new_path: Failed building data tree, xpath /ietf-interfaces:interfaces/interface[name='eth0']/ietf-ip:ipv4/mtu, libyang error 7: Not found node "mtu" in path.
Oct  2 12:19:46 infix-00-00-00 statd: ly_add_ip_mtu: Error, adding 'mtu' to data tree, libyang error 7
Oct  2 12:19:46 infix-00-00-00 statd: ly_add_ip_data: Error, adding ip MTU
Oct  2 12:19:46 infix-00-00-00 statd: ly_add_ip_addr: Error, adding ipv4 addr info for eth0
Oct  2 12:19:46 infix-00-00-00 statd: sr_ifaces_cb: Error, adding ip link info
Oct  2 12:19:46 infix-00-00-00 statd[2084]: libyang[0]: Not found node "mtu" in path. (path: Schema location "/ietf-interfaces:interfaces/interface/ietf-ip:ipv4".)
```

Since all operational data is returned, all operational functionality is broken as of v23.09.

----

**Original subject:** Unstable tests: `iface_status` and `iface_phys_address`

https://github.com/kernelkit/infix/actions/runs/6377786313/job/17307101738?pr=150#step:10:236

- iface_phys_address.py
- iface_status.py

### iface_phys_address.py

```
Starting test 0017-iface_phys_address.py
2023-10-02 09:11:26 # Starting (2023-10-02 09:11:23)
2023-10-02 09:11:26 # Probing dut1 on port d1a for IPv6LL mgmt address ...
2023-10-02 09:11:26 # Connecting to mgmt IP fe80::ff:fe00:0%d1a:830 ...
2023-10-02 09:11:26 # Target iface eth2 original mac None
2023-10-02 09:11:26 ok 1 - Initialize
2023-10-02 09:11:26 # Target iface eth2 current mac: None
2023-10-02 09:11:26 not ok 2 - Set custom MAC address
2023-10-02 09:11:26 # Exiting (2023-10-02 09:11:26)
2023-10-02 09:11:26 # Traceback (most recent call last):
[...]

PR #152, opened 2023-10-02
Add support for setting/querying IPv4/IPv6 MTU

troglobit:
This PR addresses issue #151 which broke Infix v23.09.

 - Update the minimal defconfig used by regression tests, e.g., add Python
 - Add MTU configuration support and fix IPv6 MTU support in operational data
 - Drop MTU deviation (`not-supported`), root cause of #151


rical (approved):
LGTM

PR #154, opened 2023-10-06
Fail Secure mode

troglobit:
This PR implements the *Fail Secure* mode.  Best described by this image:

fail-secure

Also included are a few measures aiming to lock down Infix, as well as fixes to some minor buglets:

 - Fix duplicate `ssh-hostkeys` and improve error handling when generating SSH keys
 - Boot order fixes by upgrading to Finit v4.5-rc5
 - Rename condition provided by sysklogd: `pid/sysklogd` -> `pid/syslogd` since the two daemons providing the service are mutually exclusive
 - Guard critical services behind the `<pid/syslogd>` condition to guarantee boot order
 - Override Finit automatically generated `dbus.conf` to ensure it always runs after udevd and any udevadm call
 - Relocate all files related to bootstrap to `src/confd/`.  The gathered files make sense to maintain together rather than scattered across the repo
 - Refactor factory-config generation to reuse scripts to generate failure-config
 - Add error case (see image above) which can be overridden by any br2-external using Infix

Failing bootstrap of YANG models, and creation of either `factory-config` or `failure-config`, now leads to the error case (VPD validation failure to error case to be added later):

fail-secure-bootstrap

Failing to load either `startup-config` or `failure-config` now leads to the error case:

fail-secure-load


src/confd/bin/gen-interfaces
```diff
@@ -0,0 +1,142 @@
+#!/bin/sh
+# Generates ietf-intefaces + ietf-ip settings for all detected interfaces
+# Interfaces must be of type Ethernet (copper/sfp), not be "internal" DSA
+# ports, and must have a "parentbus", i.e., be physical ports.
+#
+# Usage: gen-interfaces [-4bd]
+#
+#   -b brN  Bridged mode, set all interfaces as ports in brN
+#   -4      Generic option for both modes, enables ZeroConf IPv4
+#   -d      Generic option for both modes, enables DHCPv4 client
```
wkz, line 10: These options seem to only apply to the bridged mode, no?

src/confd/bin/gen-interfaces
```diff
@@ -0,0 +1,142 @@
...
+	-b)
+	    bridge="$2"
+	    shift
+	    ipv6=false		# for bridge port interfaces
+	    ;;
+	-d)
+	    dhcp=true
+	    ;;
+	-g)
+	    group="group $2"	# -g port => 'ip link show group port'
```
wkz, line 75: I think this filter should only be used to decide which subset of interfaces that should be bridged. We still want non-switch-ports to be brought up by default.

wkz (approved):
Awesome job! :rocket: 

I think we should merge this and then improve the interface generation in a followup.

src/confd/bin/gen-interfaces
troglobit: Ah yes. Will add a note of this in the refactor task to support the CRB.

src/confd/bin/gen-interfaces
troglobit: Agreed, will create a separate task to handle this.  Thanks!

PR #161, opened 2023-10-10
Alder

wkz:
Initial support for the alder board

The kernel patches are versioned here: https://github.com/kernelkit/linux/commits/kkit-linux-6.5.y/


troglobit (approved):
Very nice, great work on this! :rocket:

PR #167, opened 2023-10-13
Disable bridge multicast snooping (IGMP/MLD) by default

troglobit:
The Marvell mv88e6xxx switch driver currently used in active projects do not support multicast router ports properly yet.  Considering that Infix do not yet have YANG support for configuring IGMP/MLD snooping, we have decided to disable it in bridge setups by default.


wkz (approved):
:shipit:

mattiaswal (approved):

PR #168, opened 2023-10-16
Fix #166: drop SSDP entirely from Infix

### LINKED ISSUE #166: Drop ssdp-responder and SSDP support in Infix
The ssdp-responder in Infix refreshes it's interfaces every five seconds.  On low-powered devices, like qemu-system-aarch64 used for `make run`, this cause intermittent overloads that cause watchdog reboot in some cases.

In customer br2-externals I've disabled `ssdpd` from `factory-config` pending a local upgrade of that package in Infix.

troglobit:
mDNS-SD was added in [Windows 10 Creators Update][1] (build 1703), relased April 11, 2017.  This makes SSDP no longer critical for finding an Infix device in Windows, both ping infix-01-02-03.local and using Chrome work.

Since all major operating systems [now support mDNS-SD][2] we've decided to standardize on that and LLDP for Infix and its derivatives.  Other reasons for dropping it include, but is not limited to: lack of IPv6 in the implementation we use, and potential [security implications][3].

[1]: https://en.wikipedia.org/wiki/Windows_10,_version_1703
[2]: https://techcommunity.microsoft.com/t5/networking-blog/mdns-in-the-enterprise/ba-p/3275777
[3]: https://blog.cloudflare.com/ssdp-100gbps/


test/case/infix_services/services_basic.py
```diff
@@ -11,15 +11,13 @@
 
 import time
 import infamy
-from infamy.ssdp import SsdpClient
```
mattiaswal, line 14: Should not the infamy ssdp-module be removed as well since there are no support for it in infix left.

troglobit: The `infamy/ssdp.py` is removed at the end of the diff, I guess you missed it?

mattiaswal (commented):
Looks fine

rical (approved):
LGTM

PR #171, opened 2023-10-16
Branding, customization, and various fixes

troglobit:
Misc. branding and override support for br2-externals.


wkz (approved):
Suggest dropping 1/5 as discussed AFK. Everything else looks :fire:!

PR #180, opened 2023-10-26
Regen factory on every boot

troglobit:
This PR fixes the annoying "RMA" issue we've seen over the past week or so.  Instead of saving `factory-config` and `failure-config` to `/cfg`, we now regenerate them on each boot to match our YANG models.  The files are stored in `/run/confd/`, except for `startup-config` that remains in `/cfg`.

It also address the "unsorted interfaces" issue we found in customer bridge created by `gen-interfaces` today.

### DIFF
--- a/src/confd/bin/gen-interfaces
+++ b/src/confd/bin/gen-interfaces
@@ -85,13 +85,13 @@ phys_ifaces=$(ip -d -j link show | jq -r '
 	 select(.link_type == "ether") |
 	 select(.group != "internal")  |
 	 select(has("parentbus")) |
-	 .ifname')
+	 .ifname' | sort -V)
 ports=$(ip -d -j link show group port | jq -r '
 	 .[] |
 	 select(.link_type == "ether") |
 	 select(.group != "internal")  |
 	 select(has("parentbus")) |
-	 .ifname')
+	 .ifname' | sort -V)
 ifaces=""
 for phy in $phys_ifaces; do
     found=""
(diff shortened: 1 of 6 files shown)


src/confd/bin/gen-interfaces
```diff
@@ -85,13 +85,13 @@ phys_ifaces=$(ip -d -j link show | jq -r '
 	 select(.link_type == "ether") |
 	 select(.group != "internal")  |
 	 select(has("parentbus")) |
-	 .ifname')
+	 .ifname' | awk '{print length($0), $0}' | sort -n | cut -d' ' -f2-)
```
wkz, line 88:
```suggestion
	 .ifname' | sort -V)
```

Wouldn't this suffice? Busybox's sort implementation also comes with version sort.

src/confd/bin/gen-interfaces
```diff
@@ -85,13 +85,13 @@ phys_ifaces=$(ip -d -j link show | jq -r '
...
 ports=$(ip -d -j link show group port | jq -r '
 	 .[] |
 	 select(.link_type == "ether") |
 	 select(.group != "internal")  |
 	 select(has("parentbus")) |
-	 .ifname')
+	 .ifname' | awk '{print length($0), $0}' | sort -n | cut -d' ' -f2-)
```
wkz, line 94:
```suggestion
	 .ifname' | sort -V)
```

wkz (commented):
I think the version sort thing is worth looking into. Other than that this look :100: :+1:

src/confd/bin/gen-interfaces
troglobit: Gah, of course!  Must've been very tired last week, because for the life of me I couldn't find a working sort combination and wound up with that awk mess -- and that wasn't even the first version ... sigh

Thank you, will fix this immediately.  I'd like to get this and your PRs in for the v23.10 release.

wkz (approved):

PR #195, opened 2023-11-07
Extending the VLAN inteface documentation (picture and example)

jovatn:
- Providing an additional config example (VLAN atop brigde)
- Adding picture for this and the existing example (VLAN atop Eth) 
- Adding info on name conventions (eth.20 vs vlan10) 
- [skip ci]

Improvements on explaining the name conventions are welcome. :-)


doc/networking.md
```diff
@@ -77,7 +79,19 @@ admin@example:/config/interfaces/interface/eth0.20/> set vlan lower-layer-if eth
...
+```
+admin@example:/> configure 
+admin@example:/config/> edit interfaces interface vlan10
+admin@example:/config/interfaces/interface/vlan10/> set vlan id 10
+admin@example:/config/interfaces/interface/vlan10/> set vlan lower-layer-if br0
+admin@example:/config/interfaces/interface/vlan10/> leave
+```
+
+As conventions, a VLAN interface for VID 20 on top of an Ethernet interface *eth0* is named *eth0.20*, and a VLAN interface for VID 10 on top of a bridge interface *br0* is named *vlan10*. 
+
```
troglobit, line 93: Maybe add something like: "Remember to configure the VLAN filtering of the bridge."

troglobit (changes requested):
Can't remember rn if we have a section on bridge VLAN filtering, but we should refer to that and/or remind the user in this new section about that.

doc/networking.md
jovatn: New commits with reference to the VLAN Filtering Bridge section. In addition, that section has been extended.

troglobit (approved):

PR #200, opened 2023-11-08
Bridge VLAN filtering fixes

### LINKED ISSUE #198: VLAN membership should be explicit
We pass `vlan_default_pvid 1` when creating a bridge. It should be '0'
```
root@shelfy-82-00-00:~# cat /run/net/0/action/init/br0/50-init.ip
#!/sbin/ip -batch

link add dev br0 type bridge group_fwd_mask 0 vlan_filtering 1 mcast_snooping 0 vlan_protocol 802.1Q vlan_default_pvid 1
link set dev br0 down addrgenmode none
link set dev br0 up state up
root@shelfy-82-00-00:~#
```

The result is that VID 1 is lingering for ports and bridges in case I wish to configure a VLAN untagged for some other VID,
here VLAN 10 untagged on ports eth1 and eth2 on Bridge br0
```
admin@shelfy-82-00-00:/> show running-config 
...
interfaces {
  interface br0 {
    type bridge;
    bridge vlans {
      pvid 10;
      vlan 10 {
        untagged eth1;
        untagged eth2;
        tagged br0;
      }
    }
  }
...
  interface eth1 {
    type ethernet;
    ipv6 enabled true;
    bridge-port {
      bridge br0;
      pvid 10;
    }
  }
  interface eth2 {
    type ethernet;
    ipv6 enabled true;
    bridge-port {
      bridge br0;
      pvid 10;
    }
  }
...
  interface vlan10 {
    type vlan;
    ipv4 {
      enabled true;
      address 10.0.1.1 prefix-length 24;
    }
    vlan {
      id 10;
      lower-layer-if br0;
    }
  }
}
```
The membership looks like this
```
root@shelfy-82-00-00:~# bridge vlan show
port              vlan-id  
eth1              1 Egress Untagged
                  10 PVID Egress Untagged
eth2              1 Egress Untagged
                  10 PVID Egress Untagged
br0               1 PVID Egress Untagged
                  10
root@shelfy-82-00-00:~# 
```
Also, it does not seem to bite until after a reboot.

troglobit:
First, drop bridge vlans pvid from YANG model.  This enforces explicit port VLAN assignment, by setting `vlan_default_pvid = 0`.  Hence, untagged frames on tagged ports are now dropped by default.

Finally, fix VLAN filtering detection.  Until now the filtering was always enabled, but worked fine in tests because `vlan_default_pvid` defaulted to 1, because of the YANG model's `pvid` default value.  Instead we now check if the bridge has any VLANs configured, while there are none the bridge will be VLAN agnostic, when there is at least one we enable VLAN filtering and all bridge ports that are not a tagged or untagged member will have all its frames dropped on ingress.

### DIFF
--- a/src/confd/src/ietf-interfaces.c
+++ b/src/confd/src/ietf-interfaces.c
@@ -701,24 +701,20 @@ static const char *bridge_tagtype2str(const char *type)
 	return proto;
 }
 
-static int bridge_vlan_settings(struct lyd_node *cif, const char **proto, int *pvid)
+static int bridge_vlan_settings(struct lyd_node *cif, const char **proto)
 {
-	struct lyd_node *node;
+	struct lyd_node *vlans, *vlan;
 
-	node = lydx_get_descendant(lyd_child(cif), "bridge", "vlans", NULL);
-	if (node) {
-		const char *type = lydx_get_cattr(node, "proto");
-		const char *vid = lydx_get_cattr(node, "pvid");
+	vlans = lydx_get_descendant(lyd_child(cif), "bridge", "vlans", NULL);
+	if (vlans) {
+		const char *type = lydx_get_cattr(vlans, "proto");
+		int num = 0;
 
-		if (!type || !vid) {
-			ERROR("Missing bridge proto %s or pvid %s, defaulting to 1Q and 1.", type, vid);
-			*proto = "802.1Q";
-			*pvid = 1;
-			return 1;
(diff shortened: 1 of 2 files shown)


jovatn:
My two cents
Idea is to separate (egress) untagged and pvid settings, in order to allow for multiple (egress) untagged VLANs for a port.

wkz (approved):

PR #203, opened 2023-11-13
doc: added instructions for package override using local.mk

ahmkar94:
[skip ci]


doc/override-pckg.md
```diff
@@ -0,0 +1,44 @@
+Package override
+================
+
+Buildroot typically functions by downloading a tarball, then going through the steps of extracting, configuring, compiling, and installing the software component. The source code is unpacked into a temporary directory named output/build/\<package\>-\<version\>. This directory is ephemeral; it gets completely deleted when the command 'make clean' is executed and is recreated during the next 'make' command execution. Regardless of whether the package source code comes from a Git or Subversion repository, Buildroot first converts it into a tarball and then processes it as it would with any other tarball.
```
troglobit, line 4: Suggest using `output/build/<package>-<version>` instead, this way you don't have to escape the ><, and it gives emphasis to the path like we do elsewhere in the documentation.

This is a recurring thing below, including cases where `_` is used, which has syntactic meaning in itself in Markdown, so I suggest using back ticks for all these cases.  

Also, the (CONFIG_DIR)/local.mk is better written `$(CONFIG_DIR)/local.mk`, since `CONFIG_DIR` is a variable.

troglobit (changes requested):
Nice work!  The text is a bit rambling, I'm sure you can shorten it down, and add a reference to the Buildroot documentation instead, like we do elsewhere, for more in-depth studies.  As a general recommendation for disposition, it's always better to start with an example and explain later, or add links.

I'd also like for you to rename the file to `override-pkg.md`, or `override-package.md`, and add a (relative) link to it from https://github.com/kernelkit/infix/blob/9f792d56d671903d5c87f42a6bb2b0092d30d6c1/doc/developers-guide.md?plain=1#L78-L79

E.g., "... [using `local.mk`](override-pkg.md) ..."

Finally, we try to keep all documentation in the same style and format.  Your paragraphs are very wide.  Have a look at other documents in the same directory, each line is almost exclusively always shorter than 78 chars.  This is just a convention, but it would be great if you could address this as well.  (In Emacs we have a function called `fill-paragraph`, but it might be known in other editors as "reflow paragraph" or similar.)

troglobit (approved):
Much better, thanks! :smiley::+1:

PR #204, opened 2023-11-13
Disable root login in NETCONF builds

troglobit:
Turns out this task was simple, Buildroot already had a checkbox for it :-)

I believe this is the best default, despite our talking about enabling this (explicitly) only for release builds.  A developer can always enable this locally to debug.  This way there is no way around not using the admin user by default, even for end users.


rical:
I'm looking at the diff and all I see is the newly added row:
`# BR2_TARGET_ENABLE_ROOT_LOGIN is not set`

I was expecting to find a deleted row looking something like:
`BR2_TARGET_ENABLE_ROOT_LOGIN=y` 

What am I missing here?

troglobit:
> I'm looking at the diff and all I see is the newly added row: `# BR2_TARGET_ENABLE_ROOT_LOGIN is not set`
> 
> I was expecting to find a deleted row looking something like: `BR2_TARGET_ENABLE_ROOT_LOGIN=y`
> 
> What am I missing here?

The default is `BR2_TARGET_ENABLE_ROOT_LOGIN=y`, so to disallow root login with Kconfig syntax one has to set `# BR_TARGET_ENABLE_ROOT_LOGIN is not set` (which is what Buildroot's menuconfig does).

This gives us:

```
root@infix-00-00-00:~$ cat /etc/shadow |grep root
root:*:::::::
```

rical:
> > I'm looking at the diff and all I see is the newly added row: `# BR2_TARGET_ENABLE_ROOT_LOGIN is not set`
> > I was expecting to find a deleted row looking something like: `BR2_TARGET_ENABLE_ROOT_LOGIN=y`
> > What am I missing here?
> 
> The default is `BR2_TARGET_ENABLE_ROOT_LOGIN=y`, so to disallow root login with Kconfig syntax one has to set `# BR_TARGET_ENABLE_ROOT_LOGIN is not set` (which is what Buildroot's menuconfig does).

Right. Thanks for clearing that up. What's your take on explicitly setting to to `=n` as it's a deliberate decision that we don't want it enabled?

troglobit:
> > > I'm looking at the diff and all I see is the newly added row: `# BR2_TARGET_ENABLE_ROOT_LOGIN is not set`
> > > I was expecting to find a deleted row looking something like: `BR2_TARGET_ENABLE_ROOT_LOGIN=y`
> > > What am I missing here?
> > 
> > 
> > The default is `BR2_TARGET_ENABLE_ROOT_LOGIN=y`, so to disallow root login with Kconfig syntax one has to set `# BR_TARGET_ENABLE_ROOT_LOGIN is not set` (which is what Buildroot's menuconfig does).
> 
> Right. Thanks for clearing that up. What's your take on explicitly setting to to `=n` as it's a deliberate decision that we don't want it enabled?

I've tried that before, didn't take for some reason. Maybe because Buildroot's Kconfig is a bit older than what's in the latest kernels ... but even if `=n` worked a call to `make update-defconfig` would revert that to the commented-out syntax anyway, so no point in fighting the tide on this one.

rical (approved):
LGTM

troglobit:
AFK approved by @wkz as well :-)

PR #205, opened 2023-11-14
Alder

wkz:
More (mostly) alder-related improvements, and some generic ones:

- Encoder/decoder for ONIE formatted EEPROM's
- Define a standard LED mapping for commonly available functions and colors. Boards/externals are free to override this, of course


wkz:
Removed some remnants of an old in-kernel implementation of an `offline` trigger for `power_supply` devices.

troglobit (approved):
Very nice, I really like the LED mappings in the DTB and the way you've structured iito to work with it!

PR #206, opened 2023-11-15
Statd python backend

rical:
# Testing

- [ ] dsa interface in cli
- [ ] eth interface in cli
- [ ] veth pair in cli
- [ ] vlan in cli
- [ ] bridge in cli
- [ ] counters on hardware

# Main commit msg
```
There should be no functional change in this patch.

In this patch we rewrite the data collection in python3. This is the
new script called "yanger". It runs various commands on the system,
such as "ip" and "ethtool". It then mangles the output data from these
commands into valid YANG data.

The yanger script is started from the sysrepo callback in the statd c
code. Its output is parsed using lyd_parse_data_fd().

This means that the daemon part of statd is still written in c and the
new python code is only used when getting a query callback from
sysrepo. The c code still handles the interface netlink messages from
the kernel which controls what interfaces statd should do sysrepo
subscribe/unsubscribe to.
```


troglobit (approved):
:astonished::fire::ship:

PR #209, opened 2023-11-15
Proposed coding style for C, Linux++

troglobit:
The intention of this is to serve as a help/guidance for both devs and reviewers of pull requests.

Baseline is taken from Linux v6.5.6, with some (possibly) controversial additions (at the end of the file):

 - Do not enforce max line length
 - Alignment of variables, both when assigning and declaring
 - Use spaces to adjust when leading tabs are not enough (as Emacs)

I've also added a couple of exceptions for macros like `TAILQ_FOREACH()` which are used as control statements (if, while, for), so the opening brace should be on the same line -- otherwise clang-format thinks we are creating a recursive function ... :roll_eyes: 

This PR also contains a set of small fixes/cleanup to confd and statd.

### Emacs Users <3

Emacs users may want to set up [`format-all`](https://github.com/lassik/emacs-format-all-the-code) with `use-package`, it supports all common formatters.  For some guidance, see <https://ianyepan.github.io/posts/format-all/>.  Here's what I use myself:

```elisp
(use-package format-all
  :preface
  (defun ian/format-code ()
    "Auto-format whole buffer."
    (interactive)
    (if (derived-mode-p 'prolog-mode)
        (prolog-indent-buffer)
      (format-all-buffer)))
  :config
  ;; (global-set-key (kbd "M-F") #'ian/format-code)
  (global-set-key (kbd "M-F") #'format-all-buffer)
  (add-hook 'prog-mode-hook #'format-all-ensure-formatter))
```

Wnich can be simplified to:

```elisp
(use-package format-all
  :config
  (global-set-key (kbd "M-F") #'format-all-buffer)
  (add-hook 'prog-mode-hook #'format-all-ensure-formatter))
```

----
If you for some reason do not have `use-package` set up already, here's a quickie:

```elisp
;; Bootstrap use-package if not installed, refresh remotes, install it.
;; https://github.com/jwiegley/use-package

(require 'package)
(add-to-list 'package-archives '("gnu"   . "https://elpa.gnu.org/packages/"))
(add-to-list 'package-archives '("org"   . "http://orgmode.org/elpa/"))
(add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/"))
(package-initialize)

(unless (package-installed-p 'use-package)
  (package-refresh-contents)
  (package-install 'use-package))
(eval-and-compile
  (setq use-package-always-ensure t
	package-enable-at-startup nil
	use-package-expand-minimally t))
```


troglobit:
AFK review OK by @wkz and @mattiaswal, after removal of green goblins.

PR #210, opened 2023-11-16
board/aarch64: alder: Watchdog support

hno-addiva:


troglobit (approved):
Looks good, very clean and straight forward patch 👍

PR #213, opened 2023-11-20
CLI fixes

troglobit:
Got fed up with a few annoying things in the CLI:

 1. After latest klish upgrade, Ctrl-C to abort edit of current line stopped working
 2. `show running-config` outputs weird format compared to startup and factory
 3. Console size probing (and less prompt) does not work when klish is login shell

This PR should take care of all that.  Hopefully without causing any regressions elsewhere (I've tested manually the usual suspects and seen no issues).


package/klish/klish.conf
```diff
@@ -0,0 +1,21 @@
+# Template for config file /etc/klish/klish.conf. It's used by klish utility.
+
+# The klishd uses UNIX domain socket to receive connections. It will create an
+# filesystem entry to allow clients to find connection point. By default klish
+# client utility will connect to /tmp/klish-unix-socket.
+#UnixSocketPath=/tmp/klish-unix-socket
+
+# The klish can use external pager for non-interactive commands. By default it
+# will execute "/usr/bin/less -I -F -e -X -K -d -r -S" process as a pager.
+Pager="/usr/bin/less -I -F -e -X -K -d -r -S"
```
wkz, line 10: Should we leave it commented out, since it is the default?

wkz (approved):
nICE! :clap:

package/klish/klish.conf
troglobit: Ah, misleading comment. I've changed to -r and added -S. I'll update the comment.

troglobit: Added another commit to clean up/out the misleading comments (and move the klishd socket file to /run).

PR #214, opened 2023-11-20
`mnt` fixes

wkz:


troglobit (approved):
Nice work, this change also makes it more readable! :+1:

(Had to go to the office, have some coffee, then it was much easier to understand this :-)

PR #217, opened 2023-11-22
Add support for reading routing operational data for IPv4

mattiaswal:
In the CLI:
admin@infix-00-00-00:/> show routes
```
PREFIX                        DESTINATION                   PROTOCOL  METRIC
1.1.1.0/24                    e0                            kernel
192.168.100.0/24              1.1.1.1                       static    20
192.168.110.2/32              blackhole                     static    20
192.168.120.2/32              blackhole                     static    20
192.168.130.2/32              unreachable                   static    20
```
EDIT by @rical : framed output in markdown ```

### DIFF
--- a/board/netconf/rootfs/lib/infix/cli-pretty
+++ b/board/netconf/rootfs/lib/infix/cli-pretty
@@ -14,6 +14,12 @@ class Pad:
     state = 12
     data = 41
 
+class PadRoute:
+    prefix = 30
+    protocol = 10
+    next_hop = 30
+    metric = 10
+
 class Decore():
     @staticmethod
     def decorate(sgr, txt, restore="0"):
@@ -31,27 +37,53 @@ class Decore():
     def green(txt):
         return Decore.decorate("32", txt, "39")
 
+def get_json_data(default, indata, *args):
+    data = indata
+    for arg in args:
+        if arg in data:
+            data = data.get(arg)
+        else:
+            return default
+
+    return data
+
+class Route:
+    def __init__(self,data):
+        self.data = data
+        self.prefix = data.get('ietf-ipv4-unicast-routing:destination-prefix', '')
+        self.protocol = data.get('source-protocol','')[14:]
+        self.metric = data.get('route-preference','')
+
+        interface = get_json_data(None, self.data, 'next-hop', 'outgoing-interface')
+        address = get_json_data(None, self.data, 'next-hop', 'ietf-ipv4-unicast-routing:next-hop-address')
+        special = get_json_data(None, self.data, 'next-hop', 'special-next-hop')
+        if interface:
+            self.next_hop = interface
+        elif address:
+            self.next_hop = address
+        elif special:
+            self.next_hop = special
+        else:
+            self.next_hop = "unspecified"
+
+    def print(self):
+          row  = f"{self.prefix:<{PadRoute.prefix}}"
+          row += f"{self.next_hop:<{PadRoute.next_hop}}"
+          row += f"{self.metric:<{PadRoute.metric}}"
+          row += f"{self.protocol:<{PadRoute.protocol}}"
+          print(row)
+
 class Iface:
-    def get_json_data(self, default, *args):
-        data = self.data
-        for arg in args:
-            if arg in data:
-                data = data.get(arg)
-            else:
-                return default
-
-        return data
-
     def __init__(self, data):
         self.data = data
         self.name = data.get('name', '')
         self.type = data.get('type', '')
         self.index = data.get('if-index', '')
         self.oper_status = data.get('oper-status', '')
-        self.autoneg = self.get_json_data('unknown', 'ieee802-ethernet-interface:ethernet',
+        self.autoneg = get_json_data('unknown', self.data, 'ieee802-ethernet-interface:ethernet',
                                           'auto-negotiation', 'enable')
-        self.duplex = self.get_json_data('', 'ieee802-ethernet-interface:ethernet', 'duplex')
-        self.speed = self.get_json_data('', 'ieee802-ethernet-interface:ethernet', 'speed')
+        self.duplex = get_json_data('', self.data,'ieee802-ethernet-interface:ethernet','duplex')
+        self.speed = get_json_data('', self.data, 'ieee802-ethernet-interface:ethernet', 'speed')
         self.phys_address = data.get('phys-address', '')
 
         if data.get('statistics'):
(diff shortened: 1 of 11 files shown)


board/netconf/rootfs/lib/infix/cli-pretty
```diff
@@ -297,6 +329,24 @@ def ietf_interfaces(json, name):
...
 
+def ietf_routing(json, protocol="ipv4"):
+    if not json.get("ietf-routing:routing"):
+        print(f"Error, top level \"ietf-routing:routing\" missing")
+        sys.exit(1)
+    routes = get_json_data({}, json, 'ietf-routing:routing','ribs', 'rib')[0]
+    routes = get_json_data(None, routes, "routes", "route")
+
+    hdr = (f"{'PREFIX':<{PadRoute.prefix}}"
+           f"{'NEXT HOP':<{PadRoute.next_hop}}"
```
rical, line 340: Perhaps `NEXT-HOP` would look nicer in the header?

src/statd/statd.c
```diff
@@ -173,8 +184,8 @@ static int ly_add_yanger_data(const struct ly_ctx *ctx, struct lyd_node **parent
 }
 
 static int sr_ifaces_cb(sr_session_ctx_t *session, uint32_t, const char *path,
-			const char *, const char *, uint32_t,
-			struct lyd_node **parent, void *priv)
+                        const char *, const char *, uint32_t,
```
rical, line 187: I believe this indentation was "the kernel way"?

src/statd/statd.c
```diff
@@ -195,7 +206,38 @@ static int sr_ifaces_cb(sr_session_ctx_t *session, uint32_t, const char *path,
...
+
+static int sr_routes_cb(sr_session_ctx_t *session, uint32_t, const char *path,
+                        const char *, const char *, uint32_t,
+                        struct lyd_node **parent, __attribute__((unused)) void *priv)
+{
+	const struct ly_ctx *ctx;
+	sr_conn_ctx_t *con;
+	sr_error_t err;
+
+	ERROR("Incoming query for xpath: %s", path);
```
rical, line 226: Accidental debug left? If you really want this line I think there should be some other macro with a more suitable name

src/statd/statd.c
```diff
@@ -195,7 +206,38 @@ static int sr_ifaces_cb(sr_session_ctx_t *session, uint32_t, const char *path,
...
+
+	con = sr_session_get_connection(session);
+	if (!con) {
+		ERROR("Error, getting connection");
```
rical, line 230: I'm not sure if the function name is printed by the macro, but you could add "sr" or "sysrepo" in all these errors to make the message clearer to the reader. Such as:
`ERROR("Error, getting sr connection");`.

src/statd/statd.c
```diff
@@ -221,56 +263,46 @@ static void sr_event_cb(struct ev_loop *, struct ev_io *w, int)
 	sr_subscription_process_events(sub->sr_sub, NULL, NULL);
 }
 
-static int sub_to_iface(struct statd *statd, const char *ifname)
+static int subscribe(struct statd *statd, char *model, char *xpath, const char *name,
+                     int (*cb)(sr_session_ctx_t *session, uint32_t, const char *,
+                               const char *, const char *, uint32_t,
```
rical, line 268: Indentation problem.
I believe the kernel way is
```
int long_function_nam(int foo,
<tab><tab><tab><space>int bar)
```
Where tabs are used first and if the second line doesn't line up with the bracket it's padded with spaces.

src/statd/statd.c
```diff
@@ -221,56 +263,46 @@ static void sr_event_cb(struct ev_loop *, struct ev_io *w, int)
...
-	 * NOTE: this is a good solution but it might be to slow if a lot of
-	 * new interfaces pops up at the same time. We don't want to delay
-	 * processing the netlink messages to much.
-	 */
-	if (ip_link_check_group(ifname, "internal") == 1)
-		return SR_ERR_OK;
+	sr_error_t err;
 
-	sub = sub_find_iface(&statd->subs, ifname);
+	sub = sub_find(&statd->subs, name);
```
rical, line 275: What will happen if someone names there interface "routes"?

src/statd/statd.c
```diff
@@ -221,56 +263,46 @@ static void sr_event_cb(struct ev_loop *, struct ev_io *w, int)
...
+	snprintf(sub->name, sizeof(sub->name), "%s", name);
 
-	DEBUG("Subscribe to events for \"%s\"", path);
-	err = sr_oper_get_subscribe(statd->sr_ses, "ietf-interfaces",
-				    path, sr_ifaces_cb, sub,
-				    SR_SUBSCR_DEFAULT | SR_SUBSCR_NO_THREAD | SR_SUBSCR_DONE_ONLY,
-				    &sub->sr_sub);
+	DEBUG("Subscribe to events for \"%s\"", xpath);
+	err = sr_oper_get_subscribe(statd->sr_ses, model,
+	                            xpath, cb, sub,
```
rical, line 291: I'm not sure about the indentation

rical (commented):
Overall it looks great. I left some minor commends, mostly cosmetic. I good idea might be to run your diff though `checkpatch.pl` from the kernel.

Again, looks really good overall.

src/statd/statd.c
mattiaswal: Good point, unlikely. But i fix it.

src/statd/statd.c
mattiaswal: I just ran the proposed indentation by jocke made in #209, but i can see that it is spaces not tabs..

src/statd/statd.c
troglobit: Indeed. I've added `src/.clang-format` from the kernel, with some minor additions at the end.  It actually works.

troglobit (changes requested):
I agree with @rical's findings.

Additionally, I think I'd like to see the metric column earlier, before protocol/origin.

src/statd/statd.c
mattiaswal: But it is a function pointer? Is it still valid to do so?

troglobit: Ah, of course, didn't see that at first glance. Shame on me.

troglobit (approved):
LGTM!

Looking forward to an update in doc/networking.md, example of how to add a route and read routing table remotely using NETCONF ;-)

src/statd/statd.c
rical: Right, I missed that too. Sorry :)

PR #219, opened 2023-11-22
Bootstrap default passwd from VPD (QEMU)

rical:
## About
Use default pwd hash from VPD in QEMU.

This PR does several things. Its end goal is to fetch the admin
password hash from VPD memory during factory bootstrap.

To accomplish this probe creates a new file /run/system.json with
information read from a fw_cfg QEMU partition. The data from
/run/system.json is then later used during config bootstrap to fill in
the factory administrator password.

The idea is to make QEMU behave the same way hardware does, i.e. a
default/factory password should be fetched and used from
"hardware memory". The hardware portion of this is yet to be done.

## Future work
There's still a lot to be done in this area. The vendor specific VPD format needs to be established and the fallback flow  during bootstrap needs to be improved. The temporary "hardware fallback hash" needs to be removed and hardware support added.


troglobit (approved):
Awesome work on this! :clap: 

Edit: the regression tests don't seem too happy though?

troglobit:
@wkz maybe you can have a quick look at my additions to Richard's VPD support when you get in the office? Really hope I got gist of what we talked about during yesterdays' review.

board/common/qemu/qemu.sh
```diff
@@ -192,6 +192,38 @@ net_args()
...
+    # This is you QEMU factory/default password:
+    pwhash=$(echo -n "admin" | mkpasswd -s -m sha256crypt)
+
+    # NOTE: This format will change!
+    json_data=$(cat <<EOF
+{
+  "manufacture-date": "$(date +"%d/%m/%Y %H:%M:%S")",
+  "vendor-extension": [
+    [
+      65536,
```
wkz, line 210:
```suggestion
      61046,
```

This should use Kernelkit's IANA Enterprise Number

board/common/rootfs/lib/infix/probe
```diff
@@ -1,6 +1,47 @@
...
+    if [ $? -ne 0 ]; then
+        logger -p user.crit -t "probe" "Error, running onieprom tool"
+	return 1
+    fi
+
+    pwhash="$(echo "$vbd_json" | jq -r '.["vendor-extension"][0][1]' | jq -r '.pwhash')"
+    if [ -z "$pwhash" ] || [ "$pwhash" = "null" ]; then
+        logger -p user.crit -t "probe" "Error, didn't find password hash"
+        exit 1
+    fi
```
wkz, line 26: This is OK for now. But the real implementation needs to verify the IANA PEN

wkz (changes requested):

PR #223, opened 2023-11-23
CLI upgrade command

troglobit:
Add basic '`upgrade URI` and `show firmware` commands to CLI.


wkz (approved):
### Summary of AFK Discussion

The term "software" is a somewhat wider tent than "firmware", and should be used in cases where we must choose one term, e.g., `show software` will hopefully be able to encompass more types of blobs that the system needs than `firmware`.

troglobit:
Agreed, will update all occurrences of firmware -> software.

PR #226, opened 2023-11-26
Simplify CLI use by collapsing containers with only list elements

troglobit:
We've been annoyed by typing `edit interfaces interface e0` for a long time now, but with the addition of static routing I felt it was time to do something about it.  I started out playing with schema mounts, but since alias/subtree mounts and/or xpaths don't seem to be allowed atm. there does not seem much to be earned from going down that path[^1].  So instead I went for patching the klish-plugin-sysrepo code.

The insanity of having to type:

    admin@example:/config/> edit routing control-plane-protocols control-plane-protocol static name default static-routes ipv4
    admin@example:/config/routing/control-plane-protocols/control-plane-protocol/static/name/default/static-routes/ipv4/> set route 192.168.200.0/24 next-hop next-hop-address 192.168.1.1

has now been replaced with the slightly less annoying:

    admin@example:/config/> edit routing control-plane-protocol static name default
    admin@example:/config/routing/control-plane-protocol/static/name/default/> set ipv4 route 192.168.200.0/24 next-hop next-hop-address 192.168.1.1

The same goes for every other YANG construct where container is followed by a single list element.  E.g., `edit interface e0` instead.  For details, see https://github.com/kernelkit/klish-plugin-sysrepo/commit/fa1228a7ff2b3500b8c1bc6f6f0d28bddd2d0a91

The routing documentation and all other documentation has been updated accordingly.

----

While playing around with klish-plugin-sysrepo I also started looking at the `srp_show@` symbol and why it doesn't show everything, e.g., when in configure interface e0 context.  I couldn't fix it 100%, so that patch is for later.

[^1]: Found the following information that might be of interest later on: https://docplayer.net/234227614-Needing-an-extensible-mount-syntax-across-schema-alias-peers.html, https://github.com/sysrepo/sysrepo/issues/2842, and Clixon also describe this in their documentation today.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,16 +31,20 @@ is brief example of how to set the IP address of an interface:
 
 ```
 admin@infix-12-34-56:/> configure
-admin@infix-12-34-56:/config/> edit interfaces interface eth0
-admin@infix-12-34-56:/config/interfaces/interface/eth0/> set ipv4 <TAB>
+admin@infix-12-34-56:/config/> edit interface eth0
+admin@infix-12-34-56:/config/interface/eth0/> set ipv4 <TAB>
       address     autoconf bind-ni-name      enabled
 	  forwarding  mtu      neighbor
-admin@infix-12-34-56:/config/interfaces/interface/eth0/> set ipv4 address 192.168.2.200 prefix-length 24
-admin@infix-12-34-56:/config/interfaces/interface/eth0/> show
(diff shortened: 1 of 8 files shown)


wkz (approved):
I will miss writing `edit routing control-plane-protocols control-plane-protocol`, but I guess I'll have to live without it :wink:

PR #236, opened 2023-11-29
Fix problems when running test on real hardware not qemu

mattiaswal:
* sniffer did not work correctly
* links did not come upp as quick as in qemu
* must_reach in netns was to narrow, increase timeout
* add wait function to handle poll


wkz (approved):
:ship:

PR #237, opened 2023-11-29
confd: Don't cache libyang module references

wkz:
As it turns out, references to module objects are not safe to keep across callbacks. In particular: loading a new model into sysrepo at runtime may cause it to move objects around in memory.

Therefore, fetch a valid reference to "infix-system" on every callback instead.


troglobit (approved):
Amazing find, awesome work tracking this one down! <3

troglobit:
I was actually sniffing around this one, just had a feeling about it, but ended up spending too many hours trying to dump the schema instead.

PR #242, opened 2023-12-01
Bump to latest kkit branch for ugprades of sysrepo et al

troglobit:


wkz (approved):
Squeaky clean! :sparkles:

PR #247, opened 2023-12-08
Download required YANG models from device if missing

mattiaswal:
Also add upstream patch to make this work in netopeer2


wkz (approved):
:fire:!

Thank you sooo much for fixing this! It's been a thorn in my eye since day one.

PR #250, opened 2023-12-15
Add OSPFv2 support

mattiaswal:
A very limited part of the YANG model is implemented so far, basicly it is OSPFv2 with multiple areas and you can change timers for the interfaces. Limited operational support.

```
admin@infix-00-00-00:/config/> edit routing control-plane-protocol ietf-ospf:ospfv2 name default
admin@infix-00-00-00:/config/routing/control-plane-protocol/ietf-ospf:ospfv2/name/default/> set ospf area 0.0.0.0 interface e0 enabled true
admin@infix-00-00-00:/config/routing/control-plane-protocol/ietf-ospf:ospfv2/name/default/> leave
```


board/netconf/rootfs/libexec/infix/ospf-status
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
troglobit, line 1: Looks like you can get away with a plain `/bin/sh` heere

board/netconf/rootfs/libexec/infix/ospf-status
```diff
@@ -0,0 +1,6 @@
...
+tempfile=`mktemp`
+vtysh -c "show ip ospf interface json" | jq '[.interfaces as $interfaces | $interfaces | to_entries | group_by(.value.area)[] | {(.[0].value.area): [.[] | {name: .key} + .value] }] | reduce .[] as $item ({}; . + $item)' > $tempfile
```
troglobit, line 3: Maybe we should start writing some comments for jq expressions like this one?  "Purpose of this script is to ... "

doc/networking.md
```diff
@@ -427,6 +428,18 @@ NETCONF or using the CLI:
...
+interfaces you want to run OSPFv2.
+
+    admin@example:/config/> edit routing control-plane-protocol ietf-ospf:ospfv2 name default
+    admin@example:/config/routing/control-plane-protocol/ietf-ospf:ospfv2/name/default/> set ospf area 0.0.0.0 interface e0 enabled true
+    admin@example:/config/routing/control-plane-protocol/static/name/default/> leave
+    admin@example:/>
+
+> **Note:** The only name allowed for a control-plane-protocol is currently
+> *default*.  Meaning, you can only have one instance per routing protocol.
+
```
troglobit, line 442: I had this idea that we could add some kind of inference of the name, similar to how we guess interface type based on their names (br0/veth0), and then deviate this away?

src/confd/yang/infix-routing@2023-11-23.yang
```diff
@@ -106,7 +131,29 @@ module infix-routing {
...
+        an augmentation if there were agreement.
+
+        https://marc.info/?l=ms-ospf&m=166258444409552&w=2";
+	container redistributes {
+	  list redistribute {
+	    key "redistribute";
+	    description "Redistribute protocols into OSPF";
+	    leaf redistribute {
+	      type infix-distribute-protocol;
+	      description "Set protocol to redistribute";
```
troglobit, line 151: This feels .. odd.  How about something like:

```yang
	container redistributed {
	  list redistribute {
	    key "redistribute";
	    description "Redistribute protocols into OSPF";
	    leaf proto {
	      type infix-distribute-protocol;
	      description "Set protocol to redistribute";
```

what do you think, @wkz ?

src/statd/statd.c
```diff
@@ -256,6 +258,37 @@ static int sr_routes_cb(sr_session_ctx_t *session, uint32_t, const char *path,
...
+	if (!con) {
+		ERROR("Error, getting sr connection");
+		return SR_ERR_INTERNAL;
+	}
+
+	ctx = sr_acquire_context(con);
+	if (!ctx) {
+		ERROR("Error, acquiring context");
+		return SR_ERR_INTERNAL;
+	}
```
troglobit, line 281: You're leaking the `con` here, and I'm noticing now there are more of the same in previous functions. 

I suggest using the same pattern as in the kernel, `if (func()) { err = CODE; goto end;}`, where `end:` releases all memory and returns err?

troglobit (changes requested):

src/confd/yang/infix-routing@2023-11-23.yang
mattiaswal: Key must be proto in your example

board/netconf/rootfs/libexec/infix/ospf-status
mattiaswal: Sounds good, that script is unreadable.

doc/networking.md
mattiaswal: I have tried that. 
Name is set to key in the list, and i didn't to deviate it then.

Maybe i should give it a try again. I add it to the leftovertask

src/statd/statd.c
mattiaswal: The documentation for sr_session_get_connection say:
<pre>
Get the connection the session was created on.
</pre>

I interpret it as I should not free it in this context.

src/confd/yang/infix-routing@2023-11-23.yang
troglobit: Yeah, sorry about that.

src/statd/statd.c
troglobit: Was just an observation since you call `sr_release_context(con)` a few lines down, but not for the first error case (in all these callbacks).

src/confd/yang/infix-routing@2023-11-23.yang
mattiaswal: I fell it should be:
	container redistribute {
	  list redistribute {
	    key "proto";
	    description "Redistribute protocols into OSPF";
	    leaf proto {
	      type infix-distribute-protocol;
	      description "Set protocol to redistribute";
	      
then you can do:
admin@dut3:/config/routing/control-plane-protocol/ietf-ospf:ospfv2/name/default/> no ospf redistribute
in the CLI, which fells logical to me.

mattiaswal: Another solution is to name the container with an ending [d|s] but then the CLI will not autocomplete it when want to remove all redistribute, but that feels like a secondary issue that should be fixed inte CLI.

Maybe container redistribute[s|d] is the best approach.

troglobit (approved):

PR #253, opened 2023-12-21
Improve OSPF support

mattiaswal:
Add more operational support:
* Neighbors
* OSPF Routing table
* Router ID

Add NSSA/Stub area support


troglobit (approved):
Nice work!

PR #254, opened 2023-12-26
Classic build regression fixes

troglobit:
It's been a while since I tested the Classic builds.  Most of the regressions come from us putting NETCONF target scripts in board/common instead of board/netconf.

Maybe we should deprecate the Classic builds, or move them to another repo entirely (maybe [myLinux](https://github.com/troglobit/myLinux))?  They serve a purpose as plain/minimal embedded Linux systems, with Finit.  Useful for educational purposes and GNS3 end-devices.  But the overlap with Infix NETCONF builds is decreasing with every monthly release.  E.g., we've moved away from dhcpcd and ifupdown-ng, we use udev instead of mdev, added D-Bus, and Python is being used more and more.

What do you think, guys?


troglobit:
That was quick, thanks! :sweat_smile: 

What do you think about my question, @wkz & @mattiaswal?

wkz:
It's a tough one.

On the one hand, being able to reuse things like image formats (that fits together with our U-Boot patches), firmware upgrades, switch port discovery, etc., seems valuable - especially for smaller, special purpose systems where the overhead of sysrepo/confd might be hard to justify. One the other, I think we have to admit that keeping them around will in some ways hamper the progress of Infix proper (because it keeps breaking, as we've seen multiple times).

I would love it if we could find some way of keeping all of it in one repo, while minimizing toe-stepping.

troglobit:
@wkz thanks, and I agree. It's a really tough nut ... to improve their stability we'd need some very basic regression test, but we don't want to add to the overhead of Infix proper.

I guess one thing could be to break them out into their own repo, that use Infix as a `BR2_EXTERNAL`.  That would make it possible to reuse image formats etc., and we could use that repo as an example of the :small_red_triangle: top of the customer-value-triangle?  Not sure if it's worth the extra work right now though, so let's keep them around a while longer in the Infix tree and see what happens.

PR #255, opened 2024-01-05
Improve OSPF support

mattiaswal:
* Add BFD support (from OSPF context), no general BFD support implemented
* Add limitation to only allow one area per interface to be configured, loopback is the exception
* Show OSPF status in CLI (use vtysh for now)
* Add support for to show multiple next-hops for the same destination network in operational
* Minor bug fixes.


troglobit:
Nice progress!  Unfortunately this breaks the routing tests. Do you know why?

mattiaswal:
> Nice progress! Unfortunately this breaks the routing tests. Do you know why?

Yes, i have fixed it now, it was a problem with the test-code.

troglobit (approved):

PR #257, opened 2024-01-08
Add CLI tests for show route

mattiaswal:


troglobit:
LGTM

Just make a note of bumping `infix-routing@2023-11-23.yang` for the next PR.  We've shipped that model to end customers, so for the next release it must be a of a new version if we change it.

mattiaswal:
> LGTM
> 
> Just make a note of bumping `infix-routing@2023-11-23.yang` for the next PR. We've shipped that model to end customers, so for the next release it must be a of a new version if we change it.

No worries, it is in plan. Found a bug that I had to fix first. Last PR on OSPF support is coming today. Then we have to review the scope.

PR #260, opened 2024-01-10
Add some small but important OSPF features

mattiaswal:


troglobit (approved):
Awesome, nice work! 👍

PR #262, opened 2024-01-13
Add a larger OSPF test with multiple areas and with NSSA

mattiaswal:


troglobit:
Tests are passing now. Time to convert from draft?

mattiaswal:
> Tests are passing now. Time to convert from draft?
Need to fix one more thing first....

mattiaswal:
Now it is done, when tests pass.

troglobit (approved):
Looks **reeaally good**, nice work on all of this! <3

I think the only thing I can comment on is the lack of description in your PRs.  It would be great if you could just make a bullet list summarizing the (intended) changes in the PR -- and if it's part of a set of more PRs that complement each other.

PR #263, opened 2024-01-16
confd: add the ability to set speed, duplex and autoneg

rical:
Add the ability to configure interface speed, duplex and autoneg. Initially this was pared with YANG model deviations that restricted explicit interface speed and duplex when autoneg was enabled. But due to conflicting default states in YANG and the fact that statd also uses the same YANG meant this code was dropped.

Turning autoneg on (which is the default) is as simple as:
` > set interface x10 ethernet auto-negotiation enable true` 

However, when turning it off the kernel requires an explicit speed and duplex to be set. Ideally we would like to ensure this in the YANG model frontend, as mentioned above. This PR implements a check in the backend witch gives this error when turning off autoneg without an explicit speed and/or duplex (duplex has default):

```
> set interface x10 ethernet auto-negotiation enable false
> leave
Error: Autoneg off requires explicit speed and duplex
Please check your changes, try 'diff' and 'do show interfaces'.
```

The logic here is messy. The fact that `duplex` has a default value and `speed` does not complicates things slightly. The fact that we would prefer not to have an unused `speed` or `duplex` when autoneg is enabled is hard to solve in the backend, as we don't really want to throw an error. Also, the default `duplex` will always exists, making that check impossible without deviating it away(?)

Anyway, I tired to make the code a bit more verbose, with boilerplate like helper functions to make this messy logic easier to follow. Room for improvement here in the future.

### DIFF
--- a/board/netconf/rootfs/libexec/infix/yanger
+++ b/board/netconf/rootfs/libexec/infix/yanger
@@ -13,6 +13,9 @@ def json_get_yang_type(iface_in):
     if iface_in['link_type'] != "ether":
         return "infix-if-type:other";
 
+    if 'parentbus' in iface_in and iface_in['parentbus'] == "virtio":
+        return "infix-if-type:etherlike";
+
     if not 'linkinfo' in iface_in:
         return "infix-if-type:ethernet"
 
--- a/board/netconf/rootfs/libexec/infix/yanger
+++ b/board/netconf/rootfs/libexec/infix/yanger
@@ -584,8 +584,10 @@ if __name__ == "__main__":
 
         add_ip_link(ifname, iface_out, args.test)
         add_ip_addr(ifname, iface_out, args.test)
-        add_ethtool_groups(ifname, iface_out, args.test)
-        add_ethtool_std(ifname, iface_out, args.test)
+
+        if 'type' in iface_out and iface_out['type'] == "infix-if-type:ethernet":
+            add_ethtool_groups(ifname, iface_out, args.test)
+            add_ethtool_std(ifname, iface_out, args.test)
     elif (args.model == 'ietf-routing'):
         yang_data = {
             "ietf-routing:routing": {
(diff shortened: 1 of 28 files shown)


troglobit (approved):
I think you've made a good trade off here, so for me this gets the nod of approval.  I leave it up to @wkz to weigh in before merging though.

wkz:
I took a stab at some simplifications in e511076. Are there any corner cases I've missed there? Regardless, I think there are some nuggets in there that might be useful.

wkz (changes requested):
I took a stab at some simplifications in https://github.com/kernelkit/infix/commit/e51107606040ef6ab5986624752bc282d293efc4. Are there any corner cases I've missed there? Regardless, I think there are some nuggets in there that might be useful.

troglobit:
e511076 seems to be on a different branch than this PR originates from, but I love the simplification you made, @wkz!  I don't think it's urgent though, so lets await Monday and the return of Mr Alpe for a brief AFK discussion before merging this.

rical:
What a great improvement, thanks a lot @wkz

wkz (approved):

troglobit:
What's up with the failed jobs?

wkz:
My guess would be that `ethtool -s autoneg on e0` fails when `e0` is a `virtio_net` device.

@rical Can you have a look at this? We may need some prologue in the generated ethtool script that exits early if the interface does not support ethtool. I think `ethtool -i <IFACE>` may be a good way to test support.

troglobit:
Would it help to use the e1000 emulation in Qemu instead?

rical:
> Would it help to use the e1000 emulation in Qemu instead?
It does, but I think we would like to support virtio as well. Discussions ongoing.

rical:
This PR is blocked for now.

We need a new interface type "etherlike" to avoid trying to configure autoneg for virtio devices and such. I'm planning on picking that up next week.

troglobit:
Here we go, amazing work and a huge load of patience went into this one. Thank you, @rical! :astonished: :clinking_glasses: :partying_face:

PR #266, opened 2024-01-17
Change the default shell of the admin user to bash

wkz:
This opens up lots of scripting opportunities via ssh, without first having to install some NETCONF tools to change admin's shell. 

Some things you can now do:

- SCP works out-of-the-box now: `scp admin@[fe80::ff:fe00:0%qtap0]:/etc/motd /tmp/ix-motd`
- Config can be read/modified via SSH, using sysrepocfg: `ssh admin@fe80::ff:fe00:0%qtap0 sysrepocfg -X -fjson -d operational -x "/system/hostname"`

### DIFF
--- a/board/common/rootfs/etc/motd
+++ b/board/common/rootfs/etc/motd
@@ -1,0 +1,4 @@
+.-------.
+|  . .  | Infix -- a Network Operating System
+|-. v .-| https://kernelkit.github.io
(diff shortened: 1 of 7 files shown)


troglobit (approved):
Looks good, although that sed expression will probably haunt me to the end of my days :smile:

PR #269, opened 2024-01-18
Fix klish bug not adding prefix to identityrefs

mattiaswal:
* Update klish to fix the bug
* Update documentation (simplier configuration now)
* Add limitation (must expression) in yang to limit to supported routing protocols


troglobit (approved):
Awesome, nice work! 👍

PR #270, opened 2024-01-19
Hinting about 'help' command when starting CLI.

jovatn:
[skip ci]

### DIFF
--- a/board/netconf/rootfs/bin/clish
+++ b/board/netconf/rootfs/bin/clish
@@ -1,3 +1,5 @@
 #!/bin/sh
+# Adding hint of 'help' command when starting CLI
+echo "See the 'help' command for an introduction to the system."
 # Source settings, aliases, and probe terminal size, then hand over to klish
 exec env CLISH=yes bash -ilc /usr/bin/klish


jovatn:
The printout should probably be better formatted. Either a space or a newline. Help neeeded. :-/

```
infix-00-00-00 login: admin
Password: 
.-------.
|  . .  | Infix -- a Network Operating System
|-. v .-| https://kernelkit.github.io
'-'---'-'See the 'help' command for an introduction to the system.
admin@infix-00-00-00:/> 
```

troglobit:
Personally, I think an extra leading echo would do the trick, but I know @wkz has a sed expression in the game, so I'm leaving it to him.

wkz:
This whole mess is all down to klish not being able to interpret a literal `\n`.

I'll explain in full detail on monday.

troglobit:
Iirc, the root cause of this was that `motd` was of type string rather than binary.

My proposal is that we close this PR and add a `message-of-the-day` augment to replace it.  We can even mark `motd` as `status obsolete` to let users know that it's not implemented.  That way we can bypass the whole "how to migrate settings" issue for now.

troglobit:
Suggest closing this in favor of https://github.com/kernelkit/infix/pull/324

jovatn:
Agree! Great work!

troglobit:
Alright, closing. Please note, I've made an adjustment to the naming of the new node, inspired by openconfig.

PR #271, opened 2024-01-22
NanoPi R2S support

troglobit:
This PR adds support for the NanoPi R2S, a simple and quite powerful little two-port router.  Please see each respective commit for details.

nanopi-r2s-closed

Inside:

nanopi-r2s-open

https://wiki.friendlyelec.com/wiki/index.php/NanoPi_R2S

Specs:

  - 1 GB DDR4 RAM
  - 1x internal GbE ethernet
  - 1x GbE ethernet (using a USB3.0 to Ethernet bridge IC)
  - 1x USB2.0 port
  - Micro SD
  - micro-USB power port
  - 10x pin-header with GPIOs, I2C, IR receiver and UART


wkz (approved):
Great work!

...and cute board -- might have to pick one up :smile:

troglobit:
> Great work!

Thank you! <3

> ...and cute board -- might have to pick one up 😄

It's also ridiculously cheap as well right now, 587 SEK incl. shipping!  Probably end-of-mfg-cycle for this variant:

https://www.aliexpress.com/item/1005005596698380.html (watch out for the variant selection!!)

There are other variants as well, but not as well supported with device trees and drivers yet.

minexn:
Initial testing is working great - thank you for all your hard work.

Should we add pinout and board rate (1500000) on the Debug UART?

I noticed the LEDs on ETH1 are not working, but the network functions are normal.

troglobit:
> Initial testing is working great - thank you for all your hard work.

Great to hear! :smiley: 

> Should we add pinout and board rate (1500000) on the Debug UART?

Not following, but I was thinking of adding that to board/aarch64/r2s/README.md.  Along with how to flash, the fact that the reset button doesn't work (yet) and LED info.

> I noticed the LEDs on ETH1 are not working, but the network functions are normal.

Can confirm, seeing that here too.  Weird, could've sworn that was working at one point ... maybe related device tree, I started with Linux 5.16, went up to 6.7.1 before landing on the same kernel as base Infix use.  I'll add issues here on GitHub for it so we don't forget.

troglobit:
> I noticed the LEDs on ETH1 are not working, but the network functions are normal.

@minexn fix incoming, see https://github.com/kernelkit/infix/issues/274#issuecomment-2282700972

PR #273, opened 2024-01-22
add ietf hardware

mattiaswal:
* Fix coding style in statd.
* Reduce code duplication in statd.
* Represent the VPD as a component in ietf-hardware

### DIFF
--- a/src/statd/statd.c
+++ b/src/statd/statd.c
@@ -38,6 +38,7 @@
 #define XPATH_IFACE_BASE "/ietf-interfaces:interfaces"
 #define XPATH_ROUTING_BASE "/ietf-routing:routing/control-plane-protocols/control-plane-protocol"
 #define XPATH_ROUTING_TABLE "/ietf-routing:routing/ribs"
+#define XPATH_HARDWARE_BASE "/ietf-hardware:hardware"
 #define XPATH_ROUTING_OSPF XPATH_ROUTING_BASE "/ospf"
 
 TAILQ_HEAD(sub_head, sub);
@@ -129,7 +130,7 @@ static json_t *json_get_ip_link(void)
 }
 
 static int ly_add_yanger_data(const struct ly_ctx *ctx, struct lyd_node **parent,
-                              const char *model, const char *arg)
+			      const char *model, const char *arg)
 {
 	char *yanger_args[5] = {
 		"/libexec/infix/yanger",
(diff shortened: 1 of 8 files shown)


src/statd/statd.c
```diff
@@ -566,6 +572,12 @@ int main(int argc, char *argv[])
 		sr_disconnect(sr_conn);
 		return EXIT_FAILURE;
 	}
+	err = sub_to_hardware(&statd);
```
rical, line 575: I think it's time to make all these subscriptions calls into a function. Something like `int subscribe(&statd)` which does all these `sub_to` calls without the `sr_disconnect()` in each.

rical (commented):
Looks good. Only cosmetic comments from me.

src/statd/statd.c
mattiaswal: Yes, it is a good thing for the future to do that.

src/statd/statd.c
```diff
@@ -129,7 +129,7 @@ static json_t *json_get_ip_link(void)
 }
 
 static int ly_add_yanger_data(const struct ly_ctx *ctx, struct lyd_node **parent,
-			      char *model, const char *arg)
+                              char *model, const char *arg)
```
rical: These still look incorrect to me?

Checkpatch.pl also complains:
```
git show 211921 | ./output/build/linux-6.5.11/scripts/checkpatch.pl

ERROR: code indent should use tabs where possible                                                                                                                                        
#34: FILE: src/statd/statd.c:132:                                                                                                                                                        
+                              char *model, const char *arg)$
```

mattiaswal: i

> These still look incorrect to me?
> 
> Checkpatch.pl also complains:
> 
> ```
> git show 211921 | ./output/build/linux-6.5.11/scripts/checkpatch.pl
> 
> ERROR: code indent should use tabs where possible                                                                                                                                        
> #34: FILE: src/statd/statd.c:132:                                                                                                                                                        
> +                              char *model, const char *arg)$
> ```
That is a result of emacs linux-c-mode i think
Tab-with is set to 8 and there is 6 spaces, 

> These still look incorrect to me?
> 
> Checkpatch.pl also complains:
> 
> ```
> git show 211921 | ./output/build/linux-6.5.11/scripts/checkpatch.pl
> 
> ERROR: code indent should use tabs where possible                                                                                                                                        
> #34: FILE: src/statd/statd.c:132:                                                                                                                                                        
> +                              char *model, const char *arg)$
> ```
I used emacs with linux-c and our clang-formatter on the code... Will check the formatting script in infix.

PR #280, opened 2024-01-26
Add test to se that it is possible to get all operational data

mattiaswal:
Unlikley to fail though.


troglobit:
Aaaand it failed nevertheless! :rofl:

wkz (approved):

PR #284, opened 2024-01-29
board/aarch64: alder: Remove Micro-SD support

hno-addiva:
There is no Micro-SD slot on the alder board. Not even an optional one.


wkz (approved):

troglobit:
Tests pass, and approved, merging.

Issue #283, opened 2024-01-29
Setting hostname is adding quotation marks to the DHCP request.


minexn:
Manually setting the hostname in the DHCP request, and quotation marks are sent to the DHCP server.

<img width="248" alt="DHCP hostname" src="https://github.com/kernelkit/infix/assets/129349368/629903a1-306b-4173-9214-a8556134f440">

Config

```
set dhcp-client client-if eth0
set dhcp-client client-if eth0 option hostname value infix-r2s-1
```
This can be seen in the udhcpc process.

`udhcpc -f -p /run/dhcp-eth0.pid -t 10 -T 3 -A 10 -a1000 -S -R -x hostname:"infix-r2s-1" -i eth0 -V Infix v23.11.0-121-g051658d-dirty`

troglobit:
Nice catch!  I'm guessing -x "hostname:infix-r2s-1" would work better, right?

minexn:
Because the hostname doesn't contain spaces, we don't need quotation marks.

_RFCs mandate that a hostname's labels may contain only the ASCII letters 'a' through 'z' (case-insensitive), the digits '0' through '9', and the hyphen. Hostname labels cannot begin or end with a hyphen. No other symbols, punctuation characters, or blank spaces are permitted._

`udhcpc -f -p /run/dhcp-eth0.pid -t 10 -T 3 -A 10 -a1000 -S -R -x hostname:infix-r2s-1 -i eth0 -V Infix v23.11.0-121-g051658d-dirt`

minexn:
Another fantastic fix for infix 👍

Issue #278, opened 2024-01-24
update hostname with udhcpc script


minexn:
I do not see any hostnames assigned to the test infix clients connected to my router.

We are not sending the hostname in the udhcpc script.

 -x hostname:infix-host 

The test below works.

```
bash-5.2# udhcpc -i eth1 -x hostname:infix-host 
udhcpc: started, v1.36.0
udhcpc: broadcasting discover
udhcpc: broadcasting select for 10.10.3.216, server 10.10.3.1
udhcpc: lease of 10.10.3.216 obtained from 10.10.3.1, lease time 600
```

troglobit:
This is actually by design, at least up until now.  The primary target for Infix (currently) is switches with multiple VLANs, and since we support multiple DHCP clients it did not make sense to allow setting the hostname via DHCP.  It's a singleton, so what should we do if multiple clients try to set the name, who wins?

What we do, however, is send the current hostname to the DHCP server in case it wants to register it with the DNS server, a dnsmasq based DHCP server does this, for instance.

Is this a hard requirement for your devices to support?

(ping @wkz)

minexn:
It would be great to have both DHCP client and server support, but I see how our paths might be heading in a different direction. We are hoping to use Infix as the base for a routing platform, so we need the ability to turn off processes that might cause network issues as a gateway. I just noticed that Rapid STP is running by default. Is there a way to turn it off?

minexn:
This is the result of the PfSense DHCP request.

```
Jan 21 11:50:18 infix-r2s-x kernel: rk_gmac-dwmac ff540000.ethernet eth0: Link is Up - 1Gbps/Full - flow control off
Jan 21 11:50:18 infix-r2s-x udhcpc[535]: started, v1.36.0 
Jan 21 11:50:18 infix-r2s-x udhcpc[536]: action deconfig
Jan 21 11:50:18 infix-r2s-x finit[1]: Starting dhcp:eth0[535]
Jan 21 11:50:18 infix-r2s-x udhcpc[536]: deleting DHCP routes from eth0
Jan 21 11:50:18 infix-r2s-x udhcpc[535]: broadcasting discover 
Jan 21 11:50:21 infix-r2s-x udhcpc[535]: broadcasting discover 
Jan 21 11:50:25 infix-r2s-x last message buffered 1 times
Jan 21 11:50:25 infix-r2s-x udhcpc[535]: broadcasting select for 10.10.1.208, server 10.10.1.1 
[...]

PR #285, opened 2024-01-29
DHCP client fixes

troglobit:
This PR address the problems reported in issues #278 and #283.

### DIFF
--- a/board/common/rootfs/usr/share/udhcpc/default.script
+++ b/board/common/rootfs/usr/share/udhcpc/default.script
@@ -121,6 +121,13 @@ case "$ACTION" in
 	clr_dhcp_routes
 	set_dhcp_routes
 
+	# set hostname if given
+	if [ -n "$hostname" ]; then
+	    log "setting new hostname: $hostname"
+	    hostname "$hostname"
+	    augtool -s "set /files/etc/hosts/*[ipaddr='127.0.1.1']/canonical $hostname"
+	fi
+
 	# drop info from this interface
 	truncate -s 0 "$RESOLV_CONF"
 
--- a/board/common/rootfs/usr/share/udhcpc/default.script
+++ b/board/common/rootfs/usr/share/udhcpc/default.script
@@ -8,7 +8,7 @@ IP_CACHE="/var/lib/misc/${interface}.cache"
 RESOLV_CONF="/run/resolvconf/interfaces/${interface}.conf"
 NTPFILE="/run/chrony/dhcp-sources.d/${interface}.sources"
 [ -n "$broadcast" ] && BROADCAST="broadcast $broadcast"
-[ -n "$subnet" ] && NETMASK="$subnet"
+[ -n "$subnet" ] || subnet=32
 [ -n "$metric" ] || metric=100
 
 # Handle stateful DHCPv6 like DHCPv4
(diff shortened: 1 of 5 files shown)


src/confd/src/infix-dhcp.c
```diff
@@ -22,6 +22,25 @@ static const struct srx_module_requirement reqs[] = {
...
+        i++;
+
+    while (i < len && !isspace(arg[i]) && j < len - 1)
+        buf[j++] = buf[i++];
+
+    buf[j] = '\0';
+
+    return buf;
+}
+
```
wkz, line 43: Can't we enfore this with a `must` expression?

board/common/rootfs/usr/share/udhcpc/default.script
```diff
@@ -105,7 +105,13 @@ case "$ACTION" in
 	    /usr/sbin/avahi-autoipd -c $interface && /usr/sbin/avahi-autoipd -k $interface
 	fi
 
-	if /bin/ip addr add dev $interface $ip/$NETMASK $BROADCAST proto 5; then
+	if [ -n "$NETMASK" ]; then
+	    IP=$ip/$NETMASK
+	else
+	    IP=$ip
```
wkz, line 111:
```suggestion
	    IP=$ip/32
```
I had to try this manually (address without prefix) to see what iproute2 would do. Suggest we make it clear to future readers what will happen in this case.

board/common/rootfs/usr/share/udhcpc/default.script
```diff
@@ -105,7 +105,13 @@ case "$ACTION" in
...
+	fi
+
+	if /bin/ip addr add dev $interface $IP $BROADCAST proto 5; then
```
wkz, line 114: Do we have a ticket to fix symbolic protocol name lookups?

board/common/rootfs/usr/share/udhcpc/default.script
```diff
@@ -121,6 +121,17 @@ case "$ACTION" in
...
 	set_dhcp_routes
 
+	# set hostname if given
+	if [ -n "$hostname" ]; then
+	    log "setting new hostname: $hostname"
+	    hostname "$hostname"
+
+	    if grep -q "^127.0.1.1.*" /etc/hosts; then
+		log "Updating /etc/hosts with new hostname"
+		sed -i "s/^\(127.0.1.1\s\+\).*/\1$hostname/" /etc/hosts
```
wkz, line 131: Any utility in using `aug` here? (Like we do in `confd`)

src/confd/src/infix-dhcp.c
troglobit: This is for handling option value exceptions, like hostname, other options can have free form text string values, or hex.  It's definitely doable, but it'll take a while to figure out.

board/common/rootfs/usr/share/udhcpc/default.script
troglobit: Good point, committed your suggestion, but will probably set the default NETMASK earlier in a follow-up.

board/common/rootfs/usr/share/udhcpc/default.script
troglobit: You mean the `proto 5` ugliness? No, not that I'm aware of.

board/common/rootfs/usr/share/udhcpc/default.script
troglobit: Heh, forgot about aug. I'll have a look, that would be useful also for one of our customer cases as well.

src/confd/src/infix-dhcp.c
wkz: I got curios :angel:. This seems to work:

```yang
	  must "../name != 'hostname' or re-match(., '[a-zA-Z0-9\\-_]{0,253}')";
```

troglobit: Ha ha ha, OK thank you! I'll push another update :bowing_man::pray:

board/common/rootfs/usr/share/udhcpc/default.script
troglobit: Easy peasy with augtool :sunglasses:

board/common/rootfs/usr/share/udhcpc/default.script
troglobit: New issue #285 tracked by new project.

wkz (approved):

PR #302, opened 2024-02-13
Add possible to configure USB on/off

mattiaswal:
Requires a custom entry in the device tree, see alder.dtsi.

To see the curent state of the USB ports use netconf or CLI:
<pre>
admin@infix-00-01-00:/> show hardware 
USB PORTS           
NAME                STATE     
USB                 unlocked  
</pre>
The configuration of USB port state is done using ietf-hardware (with some additions)
<pre>
admin@infix-00-01-00:/> configure 
admin@infix-00-01-00:/config/> edit hardware 
admin@infix-00-01-00:/config/hardware/> set component USB class usb 
admin@infix-00-01-00:/config/hardware/> set component USB state admin-state locked 
admin@infix-00-01-00:/config/hardware/> leave 
admin@infix-00-01-00:/> 
</pre>

Also add USB stick on qemu machines, and a very simple test.


src/confd/src/ietf-hardware.c
```diff
@@ -0,0 +1,298 @@
...
+
+	case SR_EV_DONE:
+		/* Check if passed validation in previous event */
+		if (!fexist(SYSFS_CMDS))
+			return SR_ERR_OK;
+
+		if (systemf(SYSFS_CMDS))
+			return SR_ERR_INTERNAL;
+		(void)remove(SYSFS_CMDS);
+		return SR_ERR_OK;
```
wkz, line 220: Given that there is not really a way that `SR_EV_CHANGE` can fail, would it be easier to just return `SR_ERR_OK` from it, and then write the `authorized` files directly from confd from `SR_EV_DONE`? I.e. skip the script file and call `writedf()` from `usb_authorize()`/`dir_cb()`.

src/confd/src/ietf-hardware.c
```diff
@@ -0,0 +1,298 @@
...
+
+	case SR_EV_DONE:
+		/* Check if passed validation in previous event */
+		if (!fexist(SYSFS_CMDS))
+			return SR_ERR_OK;
+
+		if (systemf(SYSFS_CMDS))
+			return SR_ERR_INTERNAL;
+		(void)remove(SYSFS_CMDS);
```
wkz, line 219: If we do decide to keep the current construct with the script file, I think we always want to delete it here - even if `systemf()` fails.

src/confd/src/ietf-hardware.c
```diff
@@ -0,0 +1,298 @@
...
+	fputs("#!/bin/sh\n\n", fp);
+	rc = sr_get_data(session, XPATH_BASE_ "//.", 0, 0, 0, &cfg);
+	if (rc)
+		goto err_close_file;
+
+	rc = srx_get_diff(session, &diff);
+	if (rc)
+		goto err_release_data;
+
+	root = json_load_file("/run/system.json", 0, &error);
```
wkz, line 239: We parse this two times in this file. Will this show up in more places in the future? Then we might consider parsing it once and keeping it on the `struct confd`.

src/confd/yang/infix-hardware@2024-01-18.yang
```diff
@@ -25,6 +25,66 @@ module infix-hardware {
...
+
+  identity power {
+    base hardware-class;
+    description "This identity is used to describe a USB port";
+  }
+
+  identity product {
+    base hardware-class;
+    description "This identity is used to describe a USB port";
+  }
```
wkz, line 49: Copypasta detected :robot: 

Even so...do we want to represent these as 3 distinct classes, or are they all of class `vpd`?

test/case/ietf_hardware/usb.py
```diff
@@ -0,0 +1,113 @@
...
+import copy
+import infamy.hardware as hardware
+from infamy.util import until
+
+with infamy.Test() as test:
+    with test.step("Initialize"):
+        env = infamy.Env(infamy.std_topology("1x1"))
+        target = env.attach("target", "mgmt")
+        available=hardware.get_usb_ports(target)
+        assert(len(available) > 0)
```
wkz, line 13:
```suggestion
        if len(available) < 1:
            test.skip()
```

test/infamy/hardware.py
wkz, line 1: Do we think this will be used by any other tests?

If not, we might as well keep it close to the only user (i.e., in `usb.py`)

wkz (changes requested):
Great job!

wkz (approved):
Looks great! :sparkles:

troglobit (approved):
Awesome work!

PR #307, opened 2024-02-20
Add operational support for vlan filtering bridge

rical:
This PR adds operational and CLI output support for a vlan filtering bridge.

Example output:
```
INTERFACE       PROTOCOL   STATE       DATA
br0             bridge                 vlan:40u,50t                             
├ e0            bridge     DISABLED    vlan:10u,20t pvid:10                     
└ e1            bridge     BLOCKING    vlan:10t,20u pvid:20                     
br1             bridge
└ e2            bridge     FORWARDING  vlan:30u pvid:30                         
e3              ethernet   UP          02:00:00:00:00:03                        
e4              ethernet   DOWN        02:00:00:00:00:04      
```


board/netconf/rootfs/libexec/infix/cli-pretty
```diff
@@ -207,10 +212,43 @@ class Iface:
         row += f"{self.phys_address:<{Pad.data}}"
         print(row)
 
+    def pr_proto_br(self, br_vlans):
+        data_str = ""
+
+        row = f"{'bridge':<{Pad.proto}}"
+
+        if self.oper_status == "down":
+            row += Decore.red(f"{'DOWN':<{Pad.state}}")
```
wkz, line 221:
```suggestion
        if self.oper_status != "up":
            row += Decore.red(f"{self.oper_status.upper()':<{Pad.state}}")
```
I think that if the interface is not fully operational, for any reason, then I think that should have precedent over the bridge state. I.e., if the interface reports `lower-layer-down`, then that interface will never show any other bridge state than `disabled`.

board/netconf/rootfs/libexec/infix/yanger
```diff
@@ -338,6 +338,42 @@ def add_ospf(ospf):
     insert(ospf, "ietf-ospf:areas", "area", areas)
     add_ospf_routes(ospf)
 
+def get_bridge_port_pvid(ifname, iface_out, test):
+    if test:
+        cmd = ['cat', f"{test}/bridge-vlan-show-dev-{ifname}.json"]
+    else:
+        cmd = ['bridge', '-j', 'vlan', 'show', 'dev', ifname]
+
+    data = run_json_cmd(cmd)
```
wkz, line 347: This is not specific to this PR, just something that popped into my head:

Suppose we kept `test` as a global variable, and then had a single check in `run_json_cmd()` that would  translate the argument vector into a filename (e.g., to something like `{test}/bridge--j-vlan-show-dev-{ifname}.json` in this case) if `test` is set.

wkz (changes requested):
Super!

`cli-pretty` and `yanger` are a real power couple! :muscle:

board/netconf/rootfs/libexec/infix/yanger
rical: I agree. The reason for this initial implementation is that I was afraid we would end up with commands such as `foo --bar="hello"` which converted to a filename would not look pretty.

I can fix that in another PR. There's more cosmetic inconsistencies that I would like to address.

wkz (approved):
:100:

PR #319, opened 2024-02-27
statd: add good-octets counters

rical:
This PR adds out-good-octets and in-good-octets as augment in the infix-ethernet-interface yang model. These counters represents OctetsTransmittedOK and OctetsReceivedOK from ethtool.

Example CLI output:
```
root@infix-06-10-00:/> show interfaces name x10
...
eth-out-good-octets          : 4414
eth-in-good-octets           : 1897
....
```


troglobit (approved):
lgtm

PR #321, opened 2024-02-28
Fix race condition in test

mattiaswal:
There was a race condition in the test ipv4_autoconf, that appeared on slow hardware (GitHub cloud runner,  a pc with kvm disabled), that locked-up the test forever.

- also fix a little bug in infamy/iface.py


troglobit (approved):
Looks great, nice collab today 👍

Issue #328, opened 2024-03-05
PVID added on tagged port without PVID setting in configuration


mattiaswal:
```shell
admin@dut2:~$ bridge vlan
port              vlan-id  
eth1              1 PVID
eth2              1 PVID Egress Untagged
br0               1

admin@dut2:/config/> show interface eth1
type etherlike;
enabled true;
ipv6
bridge-port {
  bridge br0;
}
```

Still remains after copy running startup and reboot.

Seen when running infamy, and Infix ca44237b2883b56b14f7bcbb372, attaching full `startup-config` [startup.txt](https://github.com/kernelkit/infix/files/14495818/startup.txt)

troglobit:
Nice find!

troglobit:
> The problem only seems to exist if using VLAN id 1, changing to VLAN id 2 in the test, PVID flag was no more.

Hmm, that rings a bell. We recently changed `ietf-interfaces.c` to always create bridges with `vlan_default_pvid 0`, maybe there's a thread to pull there?

troglobit:
The root cause for this issue is that the model has "1" as default PVID.  The current code only looks at that setting to issue the PVID flag when adding the port to the VLAN.

It seems to me we need to remove the default value from the model to give the user full freedom of configuration.  E.g., tagged member in VLAN 1 but no PVID would mean untagged traffic ingressing the port would be dropped.

PR #333, opened 2024-03-07
Bridge model fixes and improved restrictions

troglobit:
## Description

This PR primarily addresses #328, and while inspecting the model the team decided to also take the time to improve the validation of port memberships in VLAN filtering bridges.

## Checklist

- [X] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [X] YANG model change => revision updated?
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Build related changes

### DIFF
--- a/src/confd/yang/infix-if-bridge@2024-03-06.yang
+++ b/src/confd/yang/infix-if-bridge@2024-03-06.yang
@@ -16,6 +16,9 @@ submodule infix-if-bridge {
   import ietf-inet-types {
     prefix inet;
   }
+  import ietf-ip {
+    prefix ip;
+  }
   import ieee802-dot1q-types {
     prefix dot1q-types;
   }
@@ -25,8 +28,17 @@ submodule infix-if-bridge {
   description  "Linux bridge extension for ietf-interfaces.";
 
   revision 2024-03-06 {
-    description "Drop default value for bridge-port PVID.
-                 Should be possible to drop untagged frames.";
+    description "Drop default value for bridge-port PVID.  Should be possible
+                 to drop untagged frames.
+
+                 Add must() expression for VLAN memberships.  Listed ports must
+                 belong to this bridge and cannot be untagged and tagged at the
+                 same time.
+
+                 Also, add must() expression to ensure VLAN filtering bridges
+                 do not have any IP address set since the bridge is always only
+                 a tagged member of VLANs.  Use a VLAN interfaces on top of the
+                 bridge for IP addressing.";
     reference "internal";
   }
   revision 2024-02-19 {
@@ -302,16 +314,39 @@ submodule infix-if-bridge {
           leaf-list untagged {
             type if:interface-ref;
             description "The set of ports in the untagged set for VLAN.";
-          }
+            must "current() = ../../../../if:name
+                  or (/if:interfaces/if:interface[if:name = current()]/bridge-port/bridge = ../../../../if:name
+                      and not(../tagged[contains(., current())]))" {
+              error-message "Port is not a member of this bridge or already set as tagged member in the same VLAN.";
+	    }
+	  }
 
           leaf-list tagged {
-            type if:interface-ref;
-            must "not(../untagged[contains(., current())])" {
-              error-message "Cannot be an untagged port at the same time.";
-            }
             description "The set of ports in the tagged set for VLAN.";
+            type if:interface-ref;
+            must "current() = ../../../../if:name
+                  or (/if:interfaces/if:interface[if:name = current()]/bridge-port/bridge = ../../../../if:name
+                      and not(../untagged[contains(., current())]))" {
+              error-message "Port is not a member of bridge or already untagged in the same VLAN.";
+            }
           }
         }
+      }
+    }
+  }
+
+  deviation "/if:interfaces/if:interface/ip:ipv4/ip:enabled" {
+    deviate add {
+      must "not(.) or count(../../infix-if:bridge/infix-if:vlans/infix-if:vlan) = 0" {
+	error-message "IPv4 address is not supported on VLAN filtering bridges.";
+      }
+    }
+  }
+
+  deviation "/if:interfaces/if:interface/ip:ipv6/ip:enabled" {
+    deviate add {
+      must "not(.) or count(../../infix-if:bridge/infix-if:vlans/infix-if:vlan) = 0" {
+	error-message "IPv6 address is not supported on VLAN filtering bridges.";
       }
     }
   }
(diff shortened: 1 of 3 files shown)


mattiaswal (approved):
:+1:

src/confd/yang/infix-if-bridge@2024-03-06.yang
```diff
@@ -297,20 +314,42 @@ submodule infix-if-bridge {
...
           }
 
           leaf-list tagged {
+            description "The set of ports in the tagged set for VLAN.";
             type if:interface-ref;
-            must "not(../untagged[contains(., current())])" {
-              error-message "Cannot be an untagged port at the same time.";
+            must "(/if:interfaces/if:interface[if:name = current()]/bridge-port/bridge = ../../../../if:name
+                   or current() = ../../../../if:name) and not(../untagged[contains(., current())])" {
+              error-message "Port not member of bridge or already set as untagged.";
```
wkz, line 327: Can you have two separate `must` expressions on the same leaf? So that the user gets a more precise error message?

src/confd/yang/infix-if-bridge@2024-03-06.yang
```diff
@@ -297,20 +314,42 @@ submodule infix-if-bridge {
...
-            description "The set of ports in the tagged set for VLAN.";
           }
         }
       }
     }
   }
 
+  augment "/if:interfaces/if:interface/ip:ipv4" {
+    leaf enabled {
+      type boolean;
```
wkz, line 337: Do we know if the `default` property from `ietf-ip` survives the augment?

src/confd/yang/infix-if-bridge@2024-03-06.yang
troglobit: Yes, all expressions separated by `and` can be placed in separate must expressions.  Good catch, fixing!

src/confd/yang/infix-if-bridge@2024-03-06.yang
troglobit: How do you do this, do you have like an intuition for these things?!  Turns out the `default` does **not survive** our augment.  I'm gonna call you :eagle: _Eagle Eye_ from now on :astonished:

src/confd/yang/infix-if-bridge@2024-03-06.yang
troglobit: Or no ... it works, technically, but breaks other functionality.

**Works:**
 - passes YANG validation with `yanglint`
 - gives better warnings for invalid configs

**Breaks:**
 - somehow confuses libyang/sysrepo so that
 - confd inference helpers stop working

For example, creating a VETH pair no longer works as we want.

**Conclusion:**

Could be a libyang bug, things start working again when I collapse two must expressions back into one.  So unfortunately the above expression has to remain as-is.

src/confd/yang/infix-if-bridge@2024-03-06.yang
troglobit: I spent a good chunk of time yesterday trying to wrap my head around this one (too).  

From our initial testing we know that we cannot just augment with a must expression, but not only does this other augment replace the ipv4/ipv6 enabled nodes with `infix-if:enabled`, which we don't want, but it also breaks default handling as you pointed out:

 - Omitting defaults does not make the new enabled nodes inherit from original node
 - Setting defaults in the model does not take, i.e., setting IPv6 infix:enabled True

Oh, and I did I mention it also does not work?  I tested back and forth various expressions but I think libyang gets confused with multiple `enabled` nodes.  So the validation doesn't run -- even after I fixed the above issues with inference.

My take away from all is that the only safe approach here is to handle this in the callback.  We could simply just log that we skipped setting any IP address since the bridge is in VLAN filtering mode.

troglobit:
See above replies to review comments.  In this latest revision I've dropped the bridge vlans vs address validation and collapsed the tagged/untagged+member expressions into a single must expression.

Still open: should we add confd callback handling to skip IP addresses on VLAN filtering bridges?

src/confd/yang/infix-if-bridge@2024-03-06.yang
troglobit: After discussions with the team today they reminded me of a little thing called "deviations", which is probably what we should try to use instead.

I'm marking this PR as draft again to finish it later.

troglobit:
Finally solved it!  The trick was to look for 'not(.)` (enabled) or `count(vlans/vlan) = 0`, otherwise, if we just look for `not(vlans)`, the expression triggers on bridges with address and no VLANs.

src/confd/yang/infix-if-bridge@2024-03-06.yang
```diff
@@ -297,20 +314,43 @@ submodule infix-if-bridge {
           leaf-list untagged {
             type if:interface-ref;
             description "The set of ports in the untagged set for VLAN.";
-          }
+            must "current() = ../../../../if:name
+                  or (/if:interfaces/if:interface[if:name = current()]/bridge-port/bridge = ../../../../if:name
+                      and not(../tagged[contains(., current())]))" {
+              error-message "Port is not a member of this bridge or already set as tagged member in the same VLAN.";
```
wkz, line 320: Splitting this up into two separate `must` expressions would maybe give better feedback to the user.

src/confd/yang/infix-if-bridge@2024-03-06.yang
```diff
@@ -297,20 +314,43 @@ submodule infix-if-bridge {
...
 
           leaf-list tagged {
+            description "The set of ports in the tagged set for VLAN.";
             type if:interface-ref;
-            must "not(../untagged[contains(., current())])" {
-              error-message "Cannot be an untagged port at the same time.";
+            must "current() = ../../../../if:name
+                  or (/if:interfaces/if:interface[if:name = current()]/bridge-port/bridge = ../../../../if:name
+                      and not(../untagged[contains(., current())]))" {
+              error-message "Port is not a member of bridge or already untagged in the same VLAN.";
```
wkz, line 330: Same comment as for untagged

wkz (approved):
Great stuff :fire:!

src/confd/yang/infix-if-bridge@2024-03-06.yang
troglobit: Yeah I tried that before, but then veth inference in the CLI stopped working.  So let's try that again after the big rebase on Buildroot 24.02 + sysrepo/libyang upgrade party.

wkz: Ahh right! My amnesia is getting worse, sorry about that! :smile:

PR #340, opened 2024-03-11
Allow bridge in factory-config for all interface types

troglobit:
This PR fixes a regression in generating factory-config for systems that want a bridge by default for all types of interfaces.


wkz (approved):
:shipit:

PR #344, opened 2024-03-14
board/aarch: styx: Fix SERDES connection to management CPU

wkz:
Only port 2 has an SGMII compatible SERDES. When referencing the SERDES (comphy), the second argument must contain the port number in order for the comphy driver to pick the correct configuration.


troglobit (approved):
lgtm 👍

PR #371, opened 2024-03-27
Misc container fixes and doc updates

troglobit:
## Description

This PR fixes a slew of container bugs found by customers.  It also adds a bit more documentation on critical topics: containers, config files, failure modes ...

## Checklist

- [X] Bugfix
  - [ ] Regression tests
- [X] Feature
  - [X] YANG model change => revision updated?
  - [ ] Regression tests added?
  - [X] Documentation added?
- [ ] Build related changes


doc/introduction.md
```diff
@@ -0,0 +1,130 @@
...
+    and already has a `startup-config`.  Provided the `startup-config`
+	is OK (see below), it is loaded and system boots successfully
+
+In both cases, external factory reset modes/button will not help, and
+in the second case will cause the device to fail on the next boot.
+
+> The second case does not yet have any warning or event that can be
+> detected from the outside.  This is planned for a later release.
+
+**Broken statrtup-config**
```
jovatn, line 104: statrtrup => startup

doc/introduction.md
```diff
@@ -0,0 +1,130 @@
...
+configuration that isolates interfaces, or even disables ports, to
+ensure that the device does not cause any security problems on the
+network.  E.g., start forwarding traffic between previously isolated
+VLANs.
+
+The generic default is isolated interfaces with communication only to
+the management CPU, SSH and console login using the device's factory
+reset password, IP connectivity only using IPv6 link-local, and device
+discovery protocols: LLDP, mDNS-SD.  The login and shell prompt are set
+to `failure-c0-ff-ee`, the last three octets of the deviice's base MAC
```
jovatn, line 123: deviice's => device's

jovatn:
Doc looks great. Found two typos, see above.
I have not yet tried the fix for adding default route or the capabiities to enable ping without privileged mode, but I will. :-)

doc/introduction.md
troglobit: Awesome, thanks!

doc/introduction.md
troglobit: Great, thank you!

troglobit:
> Doc looks great. Found two typos, see above. I have not yet tried the fix for adding default route or the capabiities to enable ping without privileged mode, but I will. :-)

You can hold off testing the build, I've found why the regression tests fail and it's a silly mistake of mine that causes confd to not start, so you would not be able to log in.  I'm working on a fix rn.

troglobit:
@wkz & @mattiaswal the runner issues turns out to have been caused by my change to confd. All that, and more, is now fixed on this branch, great pass, and it's ready for your reviews! 😎

src/confd/src/cni.c
wkz, line 1: Does the model reject routes when no address is set? Because IIUC, we now ignore configured routes if no addresses are configures.

doc/container.md
```diff
@@ -101,28 +107,186 @@ application to run.
...
+format.  Infix supports loading both `.tar` or `.tar.gz` formats.
+
+Here we show a simple example of fetching an OCI image to the system,
+but many others exist, tools like `wget`, `curl`, and `scp` come to
+mind.
+
+**Shell OCI Example:**
+
+	admin@example:~$ cd /var/tmp/
+	admin@example:/var/tmp$ sudo wget https://github.com/kernelkit/curiOS/releases/download/edge/curios-oci-amd64.tar.gzConnecting to github.com (140.82.121.3:443)
```
wkz, line 192:
```suggestion
	admin@example:/var/tmp$ sudo wget https://github.com/kernelkit/curiOS/releases/download/edge/curios-oci-amd64.tar.gz
	Connecting to github.com (140.82.121.3:443)
```

Nit: Maybe a newline got lost in the editing here?

doc/container.md
```diff
@@ -123,6 +124,139 @@ the following sections for how to add more interfaces and manage your
...
+	drwxr-xr-x    3 root     root          4096 Mar 27 14:22 ./
+	drwxr-xr-x   14 root     root          4096 Mar 27 11:57 ../
+	-rw-r--r--    1 root     root       7261785 Mar 27 14:22 curios-oci-amd64.tar.gz
+	drwx------    6 frr      frr           4096 Mar 27 11:57 frr/
+
+Importing the image into podman can be done either from the CLI
+admin-exec context ...
+
+    admin@example-c0-ff-ee:/var/tmp$ cli
+    admin@example-c0-ff-ee:/> container load /var/tmp/curios-oci-amd64.tar.gz name curios:edge
```
wkz: Not necessarily something to now, but I think we should settle on a standard of how we specify local paths. IIRC, the copy command requires a `file://` URI to copy a local file.

doc/container.md
wkz, line 1: Excellent work! :fire:

doc/introduction.md
```diff
@@ -0,0 +1,130 @@
...
+is very useful to know how to navigate and use the CLI.
+
+> Proceed to the [CLI User Guide](cli/tutorial.md).
+
+
+## Key Concepts
+
+The two modes in the CLI are the admin-exec and the configure context.
+When logging in to the system, from the console port or SSH, you land in
+admin-exec.  Here you can inspect the system status and do operations to
```
wkz, line 25: Maybe we should mention that by default, you end up in `bash`, and have to run `cli` to enter admin-exec?

doc/introduction.md
```diff
@@ -0,0 +1,130 @@
...
+
+> The [CLI Introduction](cli/introduction.md) can be useful to skim
+> through at this point.
+
+The system has several datastores (or files):
+
+ - `factory-config` consists of a set of default configurations, some
+   static and others generated per-device, e.g., a unique hostname and
+   number of ports/interfaces.   This file is generated at boot, if it
+   does not exist, i.e., only on first boot or after factory reset.
```
wkz, line 38: I think we generate factory at every boot, no?

doc/introduction.md
```diff
@@ -0,0 +1,130 @@
...
+
+After the system firmware (BIOS or and [boot loader](boot.md) start
+Linux the following happens.  The various failure modes, e.g., missing
+password in VPD, are detailed later in this section.
+
+![System boot flowchart](img/fail-secure.svg)
+
+ 1. Before mounting the `/cfg` and `/var` file systems, which host the
+    read-writable data like `startup-config` and container images, the
+	system checks if it's time for a factory reset.  If so it wipes the
```
wkz, line 65: This reads a bit like the system will do this from time to time, like `fsck` or something :smile:

Maybe something like "..., the system first checks if a factory reset has been requested by the user. If so..."

src/statd/cli-pretty
wkz, line 1: Maybe we should still print the full state name in the detailed view (`show interfaces name X`)?

I also wonder if `lower-down` or `ll-down` in the table view will be easier for a user to correlate with data they are getting via NETCONF.

src/Makefile
```diff
@@ -1,7 +1,7 @@
 # Host build of critical components, for Coverity Scan mostly
 APPS = confd execd factory keyack statd
 
-.PHONY: alla
+.PHONY: all
```
wkz, line 4: :sweden:, represent! :smile:

wkz (commented):
You're a real :superhero_man:! Great job!

src/confd/src/cni.c
troglobit: Good point, it does not, and yes the code changes in this PR are conditioned behind a static address is set.

I'll have a look at what the real limitations are and push an update.

doc/container.md
troglobit: I've tried to rid myself of $OLDJOB habits, so there's no `file://` for local files in the CLI `copy` command, and that's what I continued with here.

Ever since we switched the default shell back to Bash this has become much clearer to me.  I think we should just allow plain UNIX (absolute) paths.  The `file://` syntax has its place when YANG models or other standards require a proper [URI syntax](https://datatracker.ietf.org/doc/html/rfc3986#section-1.1), but I don't see a reason for us to follow that in Infix CLI commands.

doc/container.md
troglobit: Cheers! :clinking_glasses:

doc/introduction.md
troglobit: Great catch, fixing!

doc/introduction.md
```diff
@@ -0,0 +1,130 @@
...
+reset password, IP connectivity only using IPv6 link-local, and device
+discovery protocols: LLDP, mDNS-SD.  The login and shell prompt are set
+to `failure-c0-ff-ee`, the last three octets of the device's base MAC
+address.
+
+[^1]: YANG is a modeling language from IETF, replacing that used for
+    SNMP (MIB), used to describe the subsystems and properties of
+	the system.
+[^2]: Return Merchandise Authorization (RMA), i.e., broken beyond repair
+    by end-user and legible for return to manufacturer.
```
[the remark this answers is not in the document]
troglobit: ... aaaand I looked it up and still didn't get it right. Thanks! <3

doc/introduction.md
```diff
@@ -0,0 +1,130 @@
...
+ensure that the device does not cause any security problems on the
+network.  E.g., start forwarding traffic between previously isolated
+VLANs.
+
+The generic default is isolated interfaces with communication only to
+the management CPU, SSH and console login using the device's factory
+reset password, IP connectivity only using IPv6 link-local, and device
+discovery protocols: LLDP, mDNS-SD.  The login and shell prompt are set
+to `failure-c0-ff-ee`, the last three octets of the device's base MAC
+address.
```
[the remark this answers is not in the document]
troglobit: Good point, will try to restructure the paragraphs a bit.

src/statd/cli-pretty
troglobit: Good points, didn't consider the detailed view at all.  

Personally I favor `link-down`, in part because it's the term we use in general when we speak about port states.  I do understand where you're coming from, correlation with the base model could also be important, but I think we should lean in favor with familiarity.

doc/container.md
```diff
@@ -586,9 +587,62 @@ state data in the container's `/var/lib` is retained between reboots
...
+SSH login with keys is very handy and both remote scripting friendly
+*and secure*, but it does require a few extra configuration steps.  The
+way to set it up is covered in part in [SSH Authorized Key][4].
+
+Another *insecure* approach is to access the host system directly,
+bypassing the namespaces that make up the boundary between host and
+container.
+
+> **Security:** Please use this only in trusted setups, and possibly
+> only during a limited time frame.
```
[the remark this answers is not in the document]
troglobit: OK, I'll spice it up a bit.

doc/container.md
wkz: Agreed -- UNIX for the win! 

But ...
```
root@infix:~$ ll /cfg/
total 68
drwxr-xr-x    1 root     root          4096 Mar 28  2024 ./
drwxr-xr-x   23 root     root           344 Jan 17  2024 ../
-rw-r--r--    1 root     root         13990 Mar 28  2024 backup-config.cfg
drwxr-xr-x    4 root     root          4096 Nov 14  2103 ssl/
-rw-r--r--    1 root     root         13926 Mar 28  2024 startup-config.cfg
-rw-r--r--    1 root     root          9434 Mar 27  2024 upgrade-config.cfg
-rw-r--r--    1 root     root         13990 Mar 28  2024 wkz.cfg
root@infix:~$ cli

See the 'help' command for an introduction to the system

root@infix:/> copy /cfg/wkz.cfg running-config
Error: invalid source argument: /cfg/wkz.cfg
```
... what am I doing wrong? :smile:

doc/container.md
```diff
@@ -258,6 +262,31 @@ the upgrade command as
...
 
 
+Capabilities
+-------------
+
+An unprivileged container works for almost all use-cases, but there are
+occasions where they are too restricted and users being looking for the
+`privileged` flag.  Capabilities offers a middle ground.
+
+For example, a system container from which `ping` does not work:
```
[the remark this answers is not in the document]
wkz: Alright, good to now that we've explored the options :+1:

src/statd/cli-pretty
wkz: I see where you're coming from as well. But I think that might hide issues that we want to know about.

Unless my grep-foo fails me, the kernel will only return lower-layer-down from once [place](https://elixir.bootlin.com/linux/v6.6.22/source/net/core/link_watch.c#L56):

```c
static unsigned char default_operstate(const struct net_device *dev)
{
	if (netif_testing(dev))
		return IF_OPER_TESTING;

	/* Some uppers (DSA) have additional sources for being down, so
	 * first check whether lower is indeed the source of its down state.
	 */
	if (!netif_carrier_ok(dev)) {
		int iflink = dev_get_iflink(dev);
		struct net_device *peer;

		if (iflink == dev->ifindex)
			return IF_OPER_DOWN;

		peer = __dev_get_by_index(dev_net(dev), iflink);
		if (!peer)
			return IF_OPER_DOWN;

		return netif_carrier_ok(peer) ? IF_OPER_DOWN :
						IF_OPER_LOWERLAYERDOWN;
	}

	if (netif_dormant(dev))
		return IF_OPER_DORMANT;

	return IF_OPER_UP;
}
```

The way I read this, we should only see this state when the _all_ of the following conditions are met:
- Interface is _admin up_
- Lower interface exists is _admin up_
- Lower interface is _oper down_

In other words, this is not something that should typically show up on regular interfaces - so we want it to stand out.

troglobit: Solid argument!  I concede my point and change to `lower-down` per your previous suggestion, and restore full `lower-layer-down` for the detailed view.

src/confd/src/cni.c
troglobit: Limitation is that for routes to work with the CNI "ipam" plugin there must be at least one IP address.

I've updated the model with a must expression to validate this (Pushing update later today for 2nd review.)

doc/container.md
troglobit: Well ... even with the best of intentions, seems I screwed up that one! :sweat_smile: 

On it ... permission and ownership fixes are on another branch though.

troglobit: Fixed, works now in upcoming push!

wkz (approved):
:sparkles:!

PR #399, opened 2024-04-22
Fix USXGMII Auto-negotiation Issue

wkz:
## Description

Even when `managed = "in-band-status";` was specified on a port,
mv88e6xxx would insist on forcing the values that were supposed to be
automatically picked up by the MAC via in-band signaling.

In the case of USXGMII links on 6393X, forcing the link bit would
sometimes, for reasons that are not understood, cause the SERDES to
reset its USXGMII link partner base page (register 4.0xf0a2) to 0,
which it would then interpret as 10M/half-duplex. As a result, the
SERDES would erroneously perform rate-matching to 10M on egress
towards the PHY (which is configured for 10G/full-duplex), leading to
nothing but dropped packets and much sadness.

This can most easily be reproduced by the following setup:

```
.------------.
|        .-----.   .-------.
|        | p9  +---+ 3310P +--.
| 6393X  :-----:   :-------:  | T- or X-Units looped via external cable
|        | p10 +---+ 3310P +--'
|        '-----'   '-------'
'------------'
```
Make sure that neither p9 nor p10 are managed using in-band-status.

1. ip link set dev p9 up; ip link set dev p10 up
2. ip link set dev p9 down
3. ip link set dev p9 up

After step (3), p10 will experience the failure described above in
about 3/4 of the cases. In the cases where the issue does not arise,
simply repeat steps (2) and (3).

## Checklist

- [x] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Build related changes

### DIFF
--- a/board/aarch64/dts/alder/alder.dtsi
+++ b/board/aarch64/dts/alder/alder.dtsi
@@ -354,6 +354,7 @@
 #define XSWP(_n, _label, _mac_offs, _phy) \
 			port@_n { \
 				SWP_COMMON(_n, _label, _mac_offs, "usxgmii", _phy); \
+				managed = "in-band-status"; \
 			}
 
 #define GSWP(_n, _label, _mac_offs, _phy) \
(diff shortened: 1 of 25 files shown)


troglobit (approved):
A-M-A-Z-I-N-G work tracking this one down! :sparkles: :rocket:

PR #408, opened 2024-04-24
Support non-zero `group_fwd_mask` bridges with `mv88e6xxx` ports

wkz:
## Description

For packets with a DA in the IEEE reserved L2 group range, originating from a CPU, forward it as normal, rather than classifying it as management.

### Example Use-case:

```
     bridge (group_fwd_mask 0x4000)
     / |  \
 swp1 swp2 tap0
   \   /
(mv88e6xxx)
```

We've created a bridge with a non-zero `group_fwd_mask` (allowing LLDP in this example) containing a set of ports managed by `mv88e6xxx` and some foreign interface (e.g. an L2 VPN tunnel).

Since an LLDP packet coming in to the bridge from the other side of `tap0` is eligable for tx forward offloading, a FORWARD frame destined for `swp1` and `swp2` would be send to the conduit interface.

Before this change, due to `rsvd2cpu` being enabled on the CPU port, the switch would try to trap it back to the CPU. Given that the CPU is trusted, instead assume that it indeed meant for the packet to be forwarded like any other.

### DIFF
--- a/patches/linux/0001-net-phy-add-support-for-PHY-LEDs-polarity-modes.patch
+++ b/patches/linux/0001-net-phy-add-support-for-PHY-LEDs-polarity-modes.patch
@@ -1,7 +1,7 @@
 From 79abf7872e35080ca5b519bb4ad2acc1fbd96e8c Mon Sep 17 00:00:00 2001
 From: Christian Marangi <ansuelsmth@gmail.com>
 Date: Thu, 25 Jan 2024 21:36:59 +0100
-Subject: [PATCH 01/25] net: phy: add support for PHY LEDs polarity modes
+Subject: [PATCH 01/27] net: phy: add support for PHY LEDs polarity modes
 Organization: Addiva Elektronik
 
 Add support for PHY LEDs polarity modes. Some PHY require LED to be set
--- a/patches/linux/0002-net-mvmdio-Support-setting-the-MDC-frequency-on-XSMI.patch
+++ b/patches/linux/0002-net-mvmdio-Support-setting-the-MDC-frequency-on-XSMI.patch
@@ -1,7 +1,7 @@
 From 424ddb5d5d8ce06c578cc08cbedf519c42dd8b69 Mon Sep 17 00:00:00 2001
 From: Tobias Waldekranz <tobias@waldekranz.com>
 Date: Mon, 4 Dec 2023 11:08:11 +0100
-Subject: [PATCH 02/25] net: mvmdio: Support setting the MDC frequency on XSMI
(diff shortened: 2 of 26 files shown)


configs/aarch64_defconfig
```diff
@@ -15,7 +15,7 @@ BR2_ROOTFS_DEVICE_CREATION_DYNAMIC_EUDEV=y
 BR2_ROOTFS_DEVICE_TABLE="system/device_table.txt ${BR2_EXTERNAL_INFIX_PATH}/board/common/xattrs"
 # BR2_TARGET_ENABLE_ROOT_LOGIN is not set
 BR2_SYSTEM_BIN_SH_BASH=y
-BR2_TARGET_GENERIC_GETTY_PORT="@console"
+BR2_TARGET_GENERIC_GETTY_PORT="console"
```
troglobit, line 18: Incoming [fix](https://github.com/kernelkit/infix/commit/e11da8a338a04663e1087a21744ca59dc22755f6) to @console, on the [console-fixes](https://github.com/kernelkit/infix/commits/console-fixes/) branch, hopefully address this issue.

configs/r2s_defconfig
```diff
@@ -16,7 +16,7 @@ BR2_ROOTFS_DEVICE_CREATION_DYNAMIC_EUDEV=y
 BR2_ROOTFS_DEVICE_TABLE="system/device_table.txt ${BR2_EXTERNAL_INFIX_PATH}/board/common/xattrs ${BR2_EXTERNAL_INFIX_PATH}/board/netconf/xattrs"
 # BR2_TARGET_ENABLE_ROOT_LOGIN is not set
 BR2_SYSTEM_BIN_SH_BASH=y
-BR2_TARGET_GENERIC_GETTY_PORT="@console"
+BR2_TARGET_GENERIC_GETTY_PORT="console"
```
troglobit, line 19: Same here.

troglobit (approved):
Nice work tracking these issues down!

Only comment I have is the change to `BR2_TARGET_GENERIC_GETTY_PORT`, don't know if that was intended to be included in this PR?

troglobit:
> > Only comment I have is the change to `BR2_TARGET_GENERIC_GETTY_PORT`, don't know if that was intended to be included in this PR? To be clear, I'm 100% OK with reverting (your commit) until my `@console` branch is merged.
> I'll just wait for `console-fixes` to go in, then I'll update this PR and remove the "partial revert". Does that make sense?

Sure thing!  I mean, it's happened before that my "quick fixes" sometimes takes a few *weeks* to finalize ;-)

PR #418, opened 2024-04-29
Initial support for multiple administrators

troglobit:
## Description

Replace the hard-coded NETCONF-to-UNIX group assignment reserved for the user `admin` by instead inspecting the NACM rule set for user-to-group and group-to-acl privileges.

**Note:** unlike initially planned, no mapping of NACM groups to UNIX groups take place in this PR, only the mapping of the highest privileges to the UNIX `wheel` group is handled.  All non-admin-level users have the same permissions as before this change, with the only additional limitation that they are now confined to `/bin/clish` (or `/bin/false`) as shell.

To handle backup/restore use-cases better, users are now created with an optional UID that is read from any pre-existing home directory.  Please note, any case where the UID already is in use will cause a security trap and instead wipe the directory to be recreated by the `adduser` command.

## Other information

Also in this PR, a couple of fixes to annoying issues that crept up in this release cycle:

 * mDNS service generation was broken since 34fa6a3
 * mDNS regression test reenabled
 * Drop log level for USB "not authorized" kernel messages: err -> warning
 * Generate aux "partition" for `make run`

## Checklist

- [X] Bugfix
  - [X] Regression tests
- [X] Feature
  - [ ] Regression tests added?
  - [X] Documentation added?
- [X] Code style update (formatting, renaming)
- [ ] Build related changes


src/confd/src/ietf-system.c
```diff
@@ -1018,6 +1003,67 @@ static int change_auth(sr_session_ctx_t *session, uint32_t sub_id, const char *m
...
+				is_admin = true;
+				sr_free_values(rules, rule_count);
+			}
+			sr_free_values(groups, group_count);
+		}
+
+		if (is_admin)
+			systemf("adduser %s wheel", user);
+		else
+			systemf("delgroup %s wheel", user);
```
wkz: What is the mechanism that ensures that `user` exists here?

I.e., when loading a startup with user `foo`, how are we sure that `sys_add_user()` has already run? Conversely, when removing `foo` and its associated ACL, how are we sure that `sys_del_user()` has _not_ already run?

src/confd/src/ietf-system.c
```diff
@@ -1018,6 +1003,67 @@ static int change_auth(sr_session_ctx_t *session, uint32_t sub_id, const char *m
...
+		snprintf(xpath_groups, sizeof(xpath_groups), NACM_BASE_"/groups/group[user-name='%s']/name", user);
+		rc = sr_get_items(session, xpath_groups, 0, 0, &groups, &group_count);
+		if (!rc) {
+			for (size_t j = 0; j < group_count; j++) {
+				char xpath_rules[256];
+
+				/* Fetch and check rules for each group */
+				snprintf(xpath_rules, sizeof(xpath_rules), NACM_BASE_"/rule-list"
+					 "[group='%s']/rule[module-name='*'][access-operations='*']"
+					 "[action='permit']", groups[j].data.string_val);
```
wkz, line 1145: That's very nifty! :nerd_face:

src/confd/src/ietf-system.c
```diff
@@ -713,11 +742,39 @@ static int sys_del_user(char *user)
...
+			ERROR("Creating user %s, /home/%s existed but was owned by another uid (%d)",
+			      name, name, uid);
+			ERROR("Cleaning up stale /home/%s", name);
+			systemf("rm -rf /home/%s", name);
+			args = nargs;
+		} else {
+			/* Not in passwd or owned by us already */
+			do_chown = true;
+
+			ERROR("Reusing uid %d and /home/%s for new user %s", uid, name, name);
```
wkz, line 770: `s/ERROR/NOTE/`?

wkz (commented):
Great stuff, as always! :muscle:

board/common/rootfs/etc/passwd
```diff
@@ -0,0 +1,9 @@
+root:x:0:0:root:/root:/bin/sh
+daemon:x:1:1:daemon:/usr/sbin:/bin/false
+bin:x:2:2:bin:/bin:/bin/false
+sys:x:3:3:sys:/dev:/bin/false
+sync:x:4:100:sync:/bin:/bin/sync
+mail:x:8:8:mail:/var/spool/mail:/bin/false
+www-data:x:33:33:www-data:/var/www:/bin/false
+operator:x:37:37:Operator:/var:/bin/false
```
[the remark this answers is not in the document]
wkz: Sorry, saw that this was fixed in a followup commit.

board/common/rootfs/etc/group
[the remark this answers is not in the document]
troglobit: I've kept them as close to the original Buildroot as possible. I'm guessing there are more groups because of daemons running as nobody:$group.  But maybe we can drop `floppy`, `tape`, and `cdrom`, at least.

src/confd/src/ietf-system.c
troglobit: Sometimes the native sysrepo APIs really shine :nerd_face:

src/confd/src/ietf-system.c
troglobit: Ah yes, of course.

src/confd/src/ietf-system.c
troglobit: Yeah, there really isn't any, except for this monitor running at lower prio than the primary one used by sysrepo/netopeer2.  This is of course no guarantee to the ordering between it and ietf-system.yang. 

I considered adding a must expression to at least ensure users exist in ietf-system before adding them?  We can discuss it over a cup o' joe in the office in a bit :coffee: :croissant:

troglobit: Decisions from AFK coffee meeting:

 1. Check BusyBox impl. of `adduser` and `delgroup`, are they just manipulating strings in the `/etc/group` file, or do they verify the user exists in `/etc/passwd` as well?
 2. Check return value of `systemf()` and log error

troglobit:
Other points from our AFK coffee meeting:

 3. Remove `guest` and `operator` ACLs from our `factory-config`.  They were copy-pasted from the RFC as an example, but are completely unused and likely not relevant for any customer.
 4. We should recommend customers to read up on the NACM RFC and apply their own rule set
 5. Customers who want to protect their system should not allow non-admins even CLI access

### Example: Custom DENY Rule

Deny reading password hash for non-administrators.

```yang
nacm {
  groups {
    group admin {
      name "admin";
      user-name [ "admin1", "admin2" ];  // Specify actual administrator usernames
    }
  }

  rule-list admin-rule-list {
    group "admin";  // Link to the admin group
    rule allow-admin-password-read {
      module-name "ietf-system";
      path "/sys:system/sys:authentication/sys:user/sys:password";
      access-operations "read";  // Specify operations you want to allow (read, write, etc.)
      action permit;  // Actions can be 'permit' or 'deny'
    }
  }

  rule-list default-deny-all {
    group "*";  // Applies to all users
    rule deny-password-read {
      module-name "ietf-system";
      path "/sys:system/sys:authentication/sys:user/sys:password";
      access-operations "read";  // Specify operations you want to deny
      action deny;  // Ensuring default deny all for password field
    }
  }
}
```

board/common/rootfs/etc/group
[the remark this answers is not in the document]
troglobit: Oups, apparently not:

```
udevd[1181]: specified group 'cdrom' unknown
udevd[1181]: specified group 'tape' unknown
[ ⋯  ] Requesting device eventsudevd[1181]: specified group 'cdrom' unknown
udevd[1181]: specified group 'tape' unknown
[ OK ] 
```

I'll revert my (local) change and we'll remember to tread safely along this path.

src/confd/src/ietf-system.c
troglobit: 1. OK, so `adduser` and `delgroup` *do* check if the user exists in `/etc/passwd`.  So we need to scrap them and try again with augeas
2. Check return value of augeas

troglobit: Fixed with augeas. Dunno why I couldn't get it to work previously, but now it works at least. And great, I might add!

jovatn (approved):
Doc looks great!
Perhaps consider a rough description of what "guest" and "operator" can do, if there are any default settings for them? None? Or is it implied by the reference "primarily applies to NETCONF"?

troglobit:
> Doc looks great! Perhaps consider a rough description of what "guest" and "operator" can do, if there are any default settings for them? None? Or is it implied by the reference "primarily applies to NETCONF"?

Thanks!  We decided to drop the guest and operator (limited) acls from our `factory-config`, and direct users to the RFC instead. 

I'll update the docs out-of-band when I do the ChangeLog updates for the v24.04 release.

wkz (approved):
:sparkles:

PR #425, opened 2024-05-03
test: Improve efficiency of logical to physical topology matching

wkz:
- Present the topologies to networkx as multigraphs, rather than expaning each port of the record shape as a separate node.
- Create a topology mapping by locating a subgraph monomorphism of the logical topology in the physical ditto.

This cuts down the time of finding a mapping by several orders of magnitude. Example:

Before:

```
time python3 test/infamy/topology.py \
     	     test/virt/quad/topology.dot.in \
	     test/infamy/topologies/ring-4-duts.dot >/dev/null

real    13m1,213s
user    13m0,112s
sys     0m0,732s
```

After:

```
time python3 test/infamy/topology.py \
     	     test/virt/quad/topology.dot.in \
	     test/infamy/topologies/ring-4-duts.dot >/dev/null

real    0m0,153s
user    0m0,128s
sys     0m0,024s
```

Also, optionally use deterministic topology mappings.

Due to the salting being applied to Python's `hash()` (see [1]), the hash of an object is not stable across different Python processes. As a result, neither are the topology mappings generated by networkx.

The upside of this is randomness that we get better test coverage over time, especially on physical devices where the underlying hardware could differ between ports, for example.

On the other hand, it can be very frustrating to track down a bug when locial nodes are suffled around between phyical nodes on each invocation of a test case.

Therefore, use a random seed by default, but allow the user to specify a fixed value if they so choose. We then add a meta test that logs the seed being used for the current suite - which means we can simply copy&paste that value to rerun the suite (or single test) with the same topology mappings.

[1]: https://docs.python.org/3/reference/datamodel.html#object.__hash__

## Checklist

- [x] Bugfix
  - [ ] Regression tests
- [x] Feature
  - [x] Regression tests added?
  - [x] Documentation added?
- [ ] Build related changes

### DIFF
--- a/doc/testing.md
+++ b/doc/testing.md
@@ -410,6 +410,45 @@ You can now freely debug the network activity of your test and the
 responses from the DUT.
 
 
+### Deterministic Topology Mappings
+
+By default, mappings from logical to physical topologies are not
+stable across test case executions. This can be very frustrating when
+debugging a failing test, since logical nodes are suffled around
+between phyical nodes. In such cases, supplying a `PYTHONHASHSEED`
+variable (set to any 32-bit unsigned integer) when launching the test
+environment will make sure that topology mappings are deterministic:
+
+    $ make PYTHONHASHSEED=0 test-sh
+
+If a seed is not supplied, a random value is chosen. This seed is
+logged by the `meta/reproducible.py` test case when running a test
+suite:
+
+    $ make test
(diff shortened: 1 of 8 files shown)


troglobit (approved):
Amazing work, such a relief you managed to find the mapping issue! :sparkles:

Issue #429, opened 2024-05-08
Loading startup-config shows OK despite error detected


troglobit:
While debugging issue #428 the root cause turned out to be a system error which was _[not reported back to the console](https://github.com/kernelkit/infix/issues/428#issuecomment-2099779902)_ while loading `startup-config`.

The expected output:

```
[ OK ] Bootstrapping YANG datastore
[ OK ] Starting Configuration daemon
[ OK ] Starting DHCP/DNS proxy
[ OK ] Starting LED daemon
[ OK ] Verifying SSH host keys
[ OK ] Starting System watchdog daemon
[FAIL] Loading startup-config
[ OK ] Loading failure-config
[ OK ] Update DNS configuration
```

troglobit:
Initial analysis shows that `sysrepocfg`, from the [sysrepo project](https://www.sysrepo.org/) does not return a non-zero exit code when callbacks return a status `!= SR_ERR_OK` in the `SR_EV_DONE` event.

The documentation states:

> `SR_EV_DONE`: Occurs just after the changes have been successfully committed to the datastore, the subscriber can apply the changes now, but it cannot deny the changes in this phase anymore (any returned errors are just logged and ignored). 

We understand this, and agree that it's too late to deny any changes, yet we firmly believe that calling `sysrepocfg` should return a non-zero exit code if any callback returns non-zero.  Because when the rubber hits the road, e.g., calling iproute2 commands, or in this case *actually* calling `adduser`, we can run into all sorts of errors/bugs/issues that we cannot prepare for in `SR_EV_CHANGE`.

Issue #428, opened 2024-05-08
Cannot log in to system after upgrade to v24.04


troglobit:
After upgrade to v24.04, and v24.04.1, some users have reported they cannot log in to their system anymore.  No visible problems were logged to the console.  Here is an example reproduced in Qemu:

```
[ OK ] Bootstrapping YANG datastore
[ OK ] Starting Configuration daemon
[ OK ] Starting DHCP/DNS proxy
[ OK ] Starting LED daemon
[ OK ] Verifying SSH host keys
[ OK ] Starting System watchdog daemon
[ OK ] Loading startup-config
[ OK ] Update DNS configuration
[ OK ] Verifying self-signed https certificate
[ OK ] Starting Status daemon

Infix -- a Network Operating System v24.04.1 (hvc0)
infix-00-00-00 login: admin
Password: 
Login incorrect
```

troglobit:
Initial analysis shows a failure to add the `admin` user to the underlying Linux system:

```
May  7 15:05:35 infix confd[2407]: Reusing uid 106 and /home/admin for new user admin
May  7 15:05:35 infix confd[2407]: sys_add_new_user: Failed creating new user "admin" 
```

With some additional debug logs added:

```
May  7 16:19:12 infix confd[2407]: adduser: uid '106' in use
```

Which in turn was cause by id 106 existing in the file `/etc/group`.   Offending commit is the logic in 3f87945 which tries to map a user to an existing home directory.  What's more concerning is that `Loading startup-config` returns `[ OK ]`, this will be handled in a separate issue.

PR #431, opened 2024-05-09
Login regression and segfault fixes

troglobit:
## Description

This PR primarily deals with the following critical issues:

 - #428 
 - #429 

During the investigation several other minor, and major, issues were also found and fixed.  Please see each respective commit for details.

## Checklist

- [X] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Build related changes

### DIFF
--- a/src/confd/bin/load
+++ b/src/confd/bin/load
@@ -19,6 +19,16 @@ banner_append()
     return 0
 }
 
+note()
+{
+    logger -s -I $$ -p user.notice -t load "$*"
+}
+
+err()
+{
+    logger -s -I $$ -p user.error -t load "$*"
+}
+
 
 # shellcheck disable=SC1091
 . /etc/confdrc
@@ -35,17 +45,31 @@ else
 fi
 
 if [ ! -f "$fn" ]; then
-    logger -sik -p user.error "No such file, $fn, aborting!"
+    err "No such file, $fn, aborting!"
     exit 1
 fi
 
+note "Loading $config ..."
 if ! sysrepocfg -v3 -I"$fn" -f json; then
     case "$config" in
 	startup-config)
-	    logger -sik -p user.error "Failed loading $fn, reverting to Fail Secure mode!"
+	    err "Failed loading $fn, reverting to Fail Secure mode!"
+	    # On failure to load startup-config the system is in an undefined state
+	    cat <<-EOF >/tmp/factory.json
+		{
+		   "infix-factory-default:factory-default": {}
+		}
+		EOF
+	    # Default timeout is 2 seconds, allow time to clean up shm
+	    if ! sysrepocfg -f json -t 10 -R /tmp/factory.json; then
+		rm -f /etc/sysrepo/data/*startup*
+		rm -f /etc/sysrepo/data/*running*
+		rm -f /dev/shm/sr_*
+		killall sysrepo-plugind
+	    fi
 	    ;;
 	failure-config)
-	    logger -sik -p user.error "Failed loading $fn, aborting!"
+	    err "Failed loading $fn, aborting!"
 	    banner_append "CRITICAL ERROR: Logins are disabled, no credentials available"
 	    initctl -nbq runlevel 9
 	    ;;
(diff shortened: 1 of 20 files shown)


src/confd/bin/load
```diff
@@ -56,4 +56,7 @@ fi
 logger -sik -p user.notice "Loaded $fn successfully."
 if [ "$config" = "failure-config" ]; then
     banner_append "ERROR: Corrupt startup-config, system has reverted to default login credentials"
+else
+    # Ensure correct ownership, in particular after factory reset
+    chown admin:wheel "$fn"
```
wkz, line 86: (This might have been an issue before this change as well, but...)

Are there any guarantees that `admin` is a valid username here? Maybe `root:wheel` to reflect that it was created by the system?

src/confd/src/ietf-system.c
```diff
@@ -720,35 +693,48 @@ static int sys_del_user(char *user)
 static int sys_add_new_user(sr_session_ctx_t *sess, char *name)
```
wkz, line 693: There's quite a lot going on in this function now. Some ideas for improvements:
- Create a helper function that calls `adduser` either with or without explicit UID/GID. I.e. it would have a signature like `sys_call_adduser(sr_session_ctx_t *sess, const char *name, int uid, int gid)` where 0 could be used as sentinel value for both `uid` and `gid` since `root` will never be created. This could then also call `sys_find_usable_shell()`  to select an appropriate shell.
- Use `sys_del_user()` to remove stale user directories. This could also get rid of any SSH keys, which seems reasonable given that we've decided to nuke this user's data.
- With the two previous points in place, I think the `goto retry` falls away, which I had to read a few times before I convinced myself that there was no way of creating an infinite loop with it.

src/libsrx/src/common.h
```diff
@@ -25,6 +25,7 @@ int asprintf(char **strp, const char *fmt, ...);
 #define INFO(fmt, ...)  syslog(LOG_INFO, fmt, ##__VA_ARGS__)
 #define NOTE(fmt, ...)  syslog(LOG_NOTICE, fmt, ##__VA_ARGS__)
 #define WARN(fmt, ...)  syslog(LOG_WARNING, "%s: " fmt, __func__, ##__VA_ARGS__)
+#define ERR(fmt, ...)   syslog(LOG_ERR, fmt, ##__VA_ARGS__)
 #define ERROR(fmt, ...) syslog(LOG_ERR, "%s: " fmt, __func__, ##__VA_ARGS__)
```
wkz, line 29: It feels like we'll end up with people just copying whatever definition was closest to the code they're changing that day.

I assume this is to match the output to messages created at the info/note levels?

Maybe we should just drop `__func__` from `WARN()` and `ERROR()` instead?

wkz (commented):
Great work!

src/confd/bin/load
troglobit: Ah, good point!  As long as its got `-rw-rw---` any user in the `admin` *group* should be able to work it.  Thanks!

src/confd/src/ietf-system.c
troglobit: Wow, thanks! I thought we could've had a chat about it this morning, but you just went ahead and smashed it.  I'll get right ahead with your suggestions! :smiley:

src/libsrx/src/common.h
troglobit: Yeah these are a mess.  My take on it is that the debug macros, which need not just the correct logmask to be set but also a debug flag, they could carry the `__func__` and all warnings and errors (and notes) should be user-friendly formatted messages without function name.  Because we can usually identify the location, and when we can't (duplicate messages) we're missing line info anyway.

What do you think?

mattiaswal (approved):
Grat work! :rocket:, nice fixes.

src/libsrx/src/common.h
troglobit: I've just pushed an update with that suggestion (only `DEBUG()` has `__func__`).

src/confd/src/ietf-system.c
troglobit: Still a bit messy, but should be a lot more readable now with your suggestions in place.

src/confd/bin/load
troglobit: Fixed!

src/confd/src/ietf-system.c
wkz: Well I guess you could say its messy from all of the log messages, but we want those! Otherwise it think it looks very straightforward now! :sparkles:

wkz (approved):
All aboard! :ship:

PR #461, opened 2024-05-21
User scripts

wkz:
## Description

Add a new service that, when enabled, will execute `run-parts` on `/cfg/user-scripts.d`.

This strikes a balance between two conflicting objectives:

1. There should be no implicit way to schedule arbitrary code execution on the device, i.e. no default run-parts directory.
2. It is very useful to have a way of scheduling arbitrary code execution on the device, e.g. being able to install a debug script on a production image.

With this feature, we still meet :one:, since the feature has to be explicitly enabled in the startup-config; but we also fulfill :two:, since we can easily enable it when needed.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [x] Feature
  - [x] Regression tests added?
  - [ ] Documentation added?
- [ ] Build related changes

### DIFF
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -22,6 +22,8 @@ All notable changes to the project are documented in this file.
 - Update documentation for use of VETH pairs in containers
 - Issue #454: create bridges in `factory-config` with IGMP/MLD snooping
   enabled by default
+- Add support for optionally running user scripts from
+  `/cfg/user-scripts.d`
 
 ### Fixes
 - Add missing LICENSE hash for factory reset tool
--- a/package/skeleton-init-finit/skeleton/etc/finit.d/available/user-scripts.conf
+++ b/package/skeleton-init-finit/skeleton/etc/finit.d/available/user-scripts.conf
@@ -1,0 +1,1 @@
+task [2] run-parts /cfg/user-scripts.d -- Running user startup scripts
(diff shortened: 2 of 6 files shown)


doc/ChangeLog.md
```diff
@@ -22,6 +22,8 @@ All notable changes to the project are documented in this file.
 - Update documentation for use of VETH pairs in containers
 - Issue #454: create bridges in `factory-config` with IGMP/MLD snooping
   enabled by default
+- Add support for optionally running user scripts from
+  `/cfg/user-scripts.d`
```
troglobit, line 26: I liked the old name better, `/cfg/start.d` was short and sweet and also illustrative of when the scripts run (at startup-up).

Propose changing to that.

troglobit (changes requested):
Other than my proposed change, I see no problem with this. 

I like the warning, it should make it clear to anyone the risks involved with this.

doc/ChangeLog.md
wkz: Yeah I agree. I was trying to find a common name that would work both for the directory name and the service name.

Should we also change the name of the service in that case? Suggestions?

troglobit: Hmm, good question.  Maybe just plain _start scripts_?  

Considering the Finit [runparts](https://github.com/troglobit/finit/tree/master/doc#run-parts-scripts) supports SXXfoo and KXXbar style notation (appends `start` and `stop` to the script), maybe we want to add support for _stop scripts_ later, i.e., scripts that run at shutdown/reboot?

wkz:
After AFK discussion, we agree that we prefer to have the actual script content in the config as well.

Further discussions should happen in #463.

PR #471, opened 2024-06-02
Replace augeas with plain C and sed

troglobit:
## Description

This PR replaces the malfunctioning Augeas library with plain C APIs and a one-liner sed expression.  The reason for this fundamental change is long standing issues with Augeas, and now the latest issue where it for some scenarios fails to update `/etc/shadow` with a new password for a user.

Infix only ever used Augeas in ietf-system.c (and the DHCP helper script), so replacing the feature set "only" involves: `/etc/passwd`, `/etc/shadow`, `/etc/group`, `/etc/hostname`, and `/etc/hosts`.  For `passwd` and `shadow` the standard C APIs are used, for `group` the BusyBox tools are used, and finally, `hostname` and `hosts` are directly written with plain `fopen()` and `fputs()`.

## Checklist

- [X] Bugfix
  - [X] Regression tests
  - [X] Coverity Scan
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Build related changes


src/confd/src/ietf-system.c
```diff
@@ -849,7 +853,110 @@ static char *change_get_user(struct sr_change *change)
...
+			upw.pw_shell = (char *)shell;
+			pw = &upw;
+		}
+
+		if (putpwent(pw, fp))
+			goto fail;
+	}
+
+	/* Ensure all calls are made, if either fails we bail */
+	if (fflush(fp) | fclose(fp) | rename(_PATH_PASSWD "+", _PATH_PASSWD)) {
```
wkz, line 890:
```suggestion
	if (fclose(fp) || rename(_PATH_PASSWD "+", _PATH_PASSWD)) {
```

I don't think the bitwise OR operator makes any promises about evaluation order. But in either case `fclose(3)` says...

> The `fclose()` function flushes the stream pointed to by `stream` (writing any buffered output data using `fflush(3)`) and closes the underlying file descriptor.

...so I think we could just elide the call and rely on logical OR instead.

src/confd/src/ietf-system.c
```diff
@@ -849,7 +853,110 @@ static char *change_get_user(struct sr_change *change)
...
+			usp.sp_pwdp = lock ? "!" : (char *)hash;
+			sp = &usp;
+		}
+
+		if (putspent(sp, fp))
+			goto fail;
+	}
+
+	/* Ensure all calls are made, if either fails we bail */
+	if (fflush(fp) | fclose(fp) | rename(_PATH_SHADOW "+", _PATH_SHADOW)) {
```
wkz, line 940:
```suggestion
	if (fclose(fp) || rename(_PATH_SHADOW "+", _PATH_SHADOW)) {
```

src/confd/src/ietf-system.c
```diff
@@ -1376,41 +1397,52 @@ static int change_hostname(sr_session_ctx_t *session, uint32_t sub_id, const cha
...
-		err = SR_ERR_SYS;
+	fp = fopen(_PATH_HOSTS, "r");
+	if (!fp) {
+		err = SR_ERR_INTERNAL;
+		fclose(nfp);
 		goto err;
 	}
 
+	while (fgets(buf, sizeof(buf), fp)) {
+		ERROR("Found hosts line: %s", buf);
```
wkz, line 1435: Stray debug output?
```suggestion
```

wkz (changes requested):
Awesome! Good riddance!

src/confd/src/ietf-system.c
troglobit: Indeed, thanks!

src/confd/src/ietf-system.c
troglobit: I added this was actually after studying the BusyBox sources.  With `||` the first failing call causes subsequent ones to be skipped.  With `|` we guarantee they are all called, hence the `ffush()` in case `fclose()` for some reason fails.

wkz: Ok, but I don't think there's anything that says the compiler has to evaluate that from left to right, is there?

Also, if either the flush or close fails, don't we want to skip overwriting `/etc/passwd`? As I read it, we would be left without one now, no?

troglobit: Yeah good point, you're right!  I'll do the fixup as you suggested.

troglobit:
There, fixed.

wkz (approved):
💯

PR #475, opened 2024-06-05
QoS Improvements

wkz:
## Description

For `mv88e6xxx`: `SO_PRIORITY` can now be used to select the desired egress queue for locally originating traffic.

For VLAN interfaces, we've modeled the `{ingress,egress}-qos-map`  settings in our YANG model, so that the user can decide when to make use of priority information and when not to.

Also, fix some various bugs found along the way.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [x] Feature
  - [x] Regression tests added?
  - [ ] Documentation added?
- [ ] Build related changes


board/common/rootfs/libexec/infix/init.d/99-done
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+ln -s /run/finit/cond/reconf /run/finit/cond/usr/ixinit-done
```
troglobit, line 3: Setting conditions with `initctl` is possible early in runlevel S, and may be a cleaner way of doing this.  For an example, see:

https://github.com/kernelkit/infix/blob/91148de095b201d2ee800bfe261ac34228ea951f/board/common/rootfs/libexec/infix/mkcert#L40

Which should be `initctl -bq ...`, I'm realizing now.

troglobit (approved):
Awesome job on this, great work finding and fixing those nasty bugs too!  Only a minor suggestion on Finit conditions.

board/common/rootfs/libexec/infix/init.d/99-done
wkz: TIL, thanks! :+1:

Issue #435, opened 2024-05-13
Format specifiers for IETF system hostname and password


troglobit:
As discussed in #326, generation of parts of `factory-config` are a bit opaque to users.  An improvement would be to introduce a more declarative syntax:

- `/system/hostname`: Support `%`-specifiers. We should probably support at least:
   - `%i`: OS ID (from `/etc/os-release`). `infix` on vanilla builds,
   - `%h`: Default hostname (`DEFAULT_HOSTNAME` from `/etc/os-release`). `infix` on vanilla builds,
   - `%m`: NIC specific part of base MAC, e.g., `c0:ff:ee` translated to `c0-ff-ee`
- `/system/authentication/user/password`: Augment model to allow the literal `factory-default` or similar. This would indicate that the device's factory password, whatever it may be, should be used.

There has been requests for supporting yescrypt (#447), and similar, so updating the password hash pattern in ietf-system for `$y$` will open up for also carrying a string like `factory-default` as well.

This would make it easier to support static factory-configs.  For well-known products the interfaces and default network configuration can be hard coded.

Outstanding issue(s):
 - NETCONF SSH host key
 - OpenSSH host key (not managed by Infix currently)

The former is generated every boot for factory-config and failure-config, which is a bit unnecessary.  The `"genkey"` created is referenced by the `"default-key"` in `"ietf-netconf-server:netconf-server"`, so it could possibly be generated to a file in `/cfg/ssl/` instead at first boot, like the HTTPS certs, and sourced from there when `"genkey"` public or private key are missing, and wiped like everything else on factory reset.

Issue #308, opened 2024-02-22
Add confd version to YANG + .cfg files


troglobit:
To facilitate upgrading .cfg format we need to include a version field in YANG.  For example, we want to change our infix-system:motd from `type string` to `type binary`.

I propose using the confd version in configure.ac and including it in `infix-system-software.yang` or `infix-system.yang`.

mattiaswal:
Relates to #178

Issue #178, opened 2023-10-26
Handle breaking changes to YANG models


wkz:
In an AFK discussion, the following question was brought up:

> Sooner or later, we will have to make breaking changes to YANG models that will cause `confd` to reject old configs. How should we handle that?

Some ideas:

- `confd` should only care about the current YANG models belonging to any given version of Infix.
- We can create special text transformers that translate a valid `infix-x.y.z` config to a valid `infix-x+1.y.z` config, these can be simple scripts that run a few `jq(1)` pipelines. 
- The transformers can be stacked as infix evolves to support going from any config version to the present.
- The original config should always be preserved, so that a user can audit the diff.
- The config files need a version field, so that we know the starting point in the transformer stack.

### Config File Versioning

To keep the version information somewhat out-of-band from the actual configuration data, we should look at the possibility of adding it as a [metadata annotation](https://datatracker.ietf.org/doc/html/rfc7952#section-3.1).

Sketch:

    module infix-system {
      import ietf-yang-metadata {
        prefix "md";
      }
      md:annotation compatible {
        type string;
        description
          "Specifies the operating system which this configuration
           is compatible with. When a generated by the system
           itself, it is set to the concatenation of ID and
           VERSION_ID, as specified in /etc/os-release, joined by
           a hyphen.";
      }
    }

Example:

```js
{
    "ietf-system:system": {
        "@": {
            "infix-system:compatible": "infix-v24.06.0"
        }
    }
}
```

Having the OS name in addition to the version could potentially enable one OS to extract general settings from another, while ignoring OS-specific configuration.

### Backup/Restore

- Add `restore-config` RPC that will parse out the metadata and run the required transforms before feeding it to `sysrepo`.
- If we ever start to version config's on the device, provide a `backup-config` RPC that can fetch older versions.

troglobit:
PR https://github.com/kernelkit/infix/pull/324 shows an alternative/complementary route to take for breaking changes:

  1. Mark old setting as `deprecated` or `obsolete` (needs planning ahead)
  2. Introduce new setting replacement, used only if old setting does not exist

troglobit:
@wkz unfortunately it appears that sysrepo does not support annotations: sysrepo/sysrepo#2462 The bug report has a few years to it, but this seems to be still the case.  so I'll go ahead and add a plain config augment with a discouraging description instead.

troglobit:
Break out into own model, infix-version.yang

PR #480, opened 2024-06-16
Simplify defaults

troglobit:
## Description

This PR primarily adds support for:

 - `%h-%m` (and `%i`) hostname fmt, issue #435 
 - `$factory$` password hash, issue #435 
 - improved handling of NETCONF SSH hostkeys, issue #435 

Secondary objectives:

 - Merge infix-shell-types.yang with infix-system.yang
 - `infix-system:version`, set to confd version (1.0), issue #308
 - Migrate shell type for .cfg files of older versions, issue #178 
 - Support for static per-device factory-config
 - Minor container fixes
 - PEP-8 fixes to netconf.py

## Checklist

- [x] Bugfix
  - [ ] Regression tests
- [x] Feature
  - [x] Regression tests added?
  - [ ] Documentation added?
- [x] Code style update (formatting, renaming)
- [ ] Build related changes


test/case/ietf_system/user_admin.py
```diff
@@ -13,6 +15,47 @@
...
+        for user in users:
+            if user['name'] == 'admin':
+                admin = user
+                break
+
+        if admin is None:
+            test.fail()
+
+        if admin['password'] != "$factory$":
+            test.fail()
```
wkz, line 32: I think this means that all flavors of Infix now has to have an admin user with the factory password in its default config, which does not seem quite right.

Maybe the test could be to add another user (`jacky`/`$factory$`), and verify that we can login as that user, with the device's default password (taken from the physical topology)? That would also avoid the test having to know about `system.json`.

test/case/ietf_system/hostname.py
```diff
@@ -1,13 +1,16 @@
...
+import random
+import string
+import re
 import infamy
 
 with infamy.Test() as test:
     with test.step("Initialize"):
         env = infamy.Env(infamy.std_topology("1x1"))
         target = env.attach("target", "mgmt")
+        tgt = env.attach("target", "mgmt", "ssh")
```
wkz, line 12: I think we should settle on a convention for these names. The user test used `tgtssh` for the SSH channel. Maybe there's something even better, but I think that `target` and `tgt` may be confusing to new readers.

wkz (changes requested):
I envy your productivity, great job! :1st_place_medal:

src/confd/src/ietf-system.c
```diff
@@ -1237,34 +1272,22 @@ static int change_nacm(sr_session_ctx_t *session, uint32_t sub_id, const char *m
...
 		if (set_shell(user, shell))
 			ERROR("Failed adjusting shell for user %s", user);
+
+		if (is_admin)
+			add_group(user, "wheel");
+		else
+			del_group(user, "wheel");
+
+		/* Users with a shell are allowed CLI access */
+		if (strcmp(shell, "false"))
```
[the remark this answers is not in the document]
troglobit: Indeed, dunno what I thinking here.  Although, when I just tested this, only foo user have shell set to false.  I get:

```
admin:x:1000:1000:Linux User,,,:/home/admin:/bin/bash
foo:x:1001:1001:Linux User,,,:/home/foo:/bin/false
guest:x:1002:1002:Linux User,,,:/home/guest:/bin/clish
```

Which flies in the face of everything I thought I knew about `strcmp()` ... I'll try yo get to the bottom of this :rofl:

src/confd/src/ietf-system.c
```diff
@@ -1375,29 +1445,132 @@ static int change_editor(sr_session_ctx_t *session, uint32_t sub_id, const char
...
 }
 
+static char *get_mac(struct confd *confd, char *mac, size_t len)
+{
+	struct json_t *obj;
+
+	obj = json_object_get(confd->root, "mac-address");
+	if (!json_is_string(obj)) {
+	fallback:
+		snprintf(mac, len, "c0-ff-ee");
```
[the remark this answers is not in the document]
troglobit: Good point, that's obviously better.

src/confd/src/ietf-system.c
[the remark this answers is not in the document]
troglobit: Indeed, thanks fixing!

test/case/ietf_system/hostname.py
troglobit: OK, fixing.

src/confd/src/ietf-system.c
```diff
@@ -1416,29 +1420,132 @@ static int change_editor(sr_session_ctx_t *session, uint32_t sub_id, const char
...
+ *
+ * %i: OS ID (from /etc/os-release). infix on vanilla builds,
+ * %h: Default hostname (DEFAULT_HOSTNAME from /etc/os-release). infix on vanilla builds,
+ * %m: NIC specific part of base MAC, e.g., c0:ff:ee translated to c0-ff-ee
+ * %%: Literal %
+ *
+ * NOTE: to be forward compatible, any unknown % combination will be silently consumed.
+ *       E.g., "example-%z" will become "example-"
+ */
+static int hostnamefmt(struct confd *confd, char **fmt)
```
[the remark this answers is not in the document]
troglobit: Sure, for any future extension that's a good idea.  I'll put that as a big comment.

utils/sysrepo-load-modules.sh
```diff
@@ -48,8 +41,8 @@ INSTALL_MODULE_CMD() {
 }
 
 UPDATE_MODULE() {
-	CMD="'$SYSREPOCTL' -U $MODDIR/$1  -s '$MODDIR -v2"
-    eval $CMD
+    CMD="'$SYSREPOCTL' -U $MODDIR/$1  -s '$MODDIR' -v2"
```
mattiaswal, line 44: This breaks relations with the original script.  Everything below source - is copied from netopeer2

mattiaswal (changes requested):
Massive work, :+1:  some small things.

board/common/rootfs/libexec/infix/init.d/30-cfg-migrate
```diff
@@ -0,0 +1,102 @@
...
+
+err()
+{
+    logger -k -p user.err -t "$ident" "$1"
+}
+
+file_version()
+{
+    jq -r '
+        if .["ietf-system:system"] | has("version") then
```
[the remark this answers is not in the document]
troglobit: Hmm, but I fixed that this morning ... wtf?  Thanks for finding this!!

utils/sysrepo-load-modules.sh
troglobit: Yeah well update didn't work because of the missing closing ' after $MODDIR, so confd-rebuild never worked.  I think we should break with netopeer2 anyway.

mattiaswal: Strange, works for me.

board/common/rootfs/libexec/infix/init.d/30-cfg-migrate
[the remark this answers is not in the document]
troglobit: Fixed.

troglobit:
There, final comments and regressions fixed!  Tests pass locally.

troglobit:
Aaand now tests also pass on the server. Best solstice gift ever! :sun_with_face: :partying_face:

mattiaswal (approved):
LGTM Great work :+1:

src/confd/src/ietf-keystore.c
```diff
@@ -0,0 +1,135 @@
...
+	return st.st_size;
+}
+
+/* sanity check, must exist and be of non-zero size */
+static bool fileok(const char *fn)
+{
+	if (!fexist(fn))
+		return false;
+
+	if (filesz(fn) < 42)
```
wkz, line 29: Maybe a comment about why this length was chosen?

src/confd/src/ietf-keystore.c
```diff
@@ -0,0 +1,135 @@
...
+		return false;
+
+	return true;
+}
+
+static char *filerd(const char *fn, size_t len)
```
wkz, line 35:
```suggestion
static char *key_from_pem(const char *fn, size_t len)
```
I had to read this many times before I was able to parse "file read" (I think). Maybe something a bit more descriptive?

wkz (approved):
Awesome work! Sorry about the review lag.

src/confd/src/ietf-keystore.c
troglobit: Fixed, thanks!

src/confd/src/ietf-keystore.c
troglobit: Thanks, added comments to all helper functions.

PR #485, opened 2024-06-18
Refactor Confd Yang Modules Setup Script

ahmkar94:
These changes aim to improve the modularity and maintainability of our codebase, making it easier to manage and extend the YANG module setup process in the future.


utils/yang_modules_setup.sh
```diff
@@ -0,0 +1,164 @@
...
+    fi
+}
+
+MODULES_LIST_FILE_CONFD=${MODULES_LIST_FILE:-$(dirname "$0")/yang_modules_confd.inc}
+MODULES_LIST_FILE_CONTAINTERS=${MODULES_LIST_FILE:-$(dirname "$0")/yang_modules_containers.inc}
+MODULES_LIST_FILE_INTERFACES=${MODULES_LIST_FILE:-$(dirname "$0")/yang_modules_interfaces.inc}
+
+check_file_exists $MODULES_LIST_FILE_CONFD
+check_file_exists $MODULES_LIST_FILE_CONTAINTERS
+check_file_exists $MODULES_LIST_FILE_INTERFACES
```
mattiaswal, line 26: The file should be as an input to the script example:
```
./yang_modules_setup.sh file-with-modules
```
And in confd, just call the script twice
```
yang_modules_setup.sh confd.inc
yang_modules_setup.sh interfaces.inc
```

utils/yang_modules_setup.sh
```diff
@@ -0,0 +1,164 @@
...
+check_file_exists() {
+    file=$1
+
+    if [ ! -f "$file" ]; then
+        echo "Modules list file not found: $file"
+        exit 1
+    fi
+}
+
+MODULES_LIST_FILE_CONFD=${MODULES_LIST_FILE:-$(dirname "$0")/yang_modules_confd.inc}
```
mattiaswal, line 20: Didn't it work to have a variable called MODULES as before and use source on the file requested?

mattiaswal (changes requested):

mattiaswal (approved):
Great work :fireworks:  :+1:

troglobit:
I'm not really happy with all the whitespace changes in this PR.  Lots of initial indent introduced in the .mk file that does not comply with Buildroot standards.

Also, I'm left a confounded with the move of logic from src/confd to package/confd.  Why can't we keep the internals of which modules are needed in src/confd?

mattiaswal:
> I'm not really happy with all the whitespace changes in this PR. Lots of initial indent introduced in the .mk file that does not comply with Buildroot standards.
> 
> Also, I'm left a confounded with the move of logic from src/confd to package/confd. Why can't we keep the internals of which modules are needed in src/confd?

The logic, what do you mean? The script itself has been moved to utils/ to make it nicer when calling from an external build, also remove all knowledge about containers etc from that feels right, with that in mind. The calling of the script has always been in confd.mk. Is it the .inc files you do not like in src/confd/, I can agree, it should be moved, it should be with the source code.

ahmkar94:
I had a feeling that I shouldn't touch the indentations. I just wasn't used to see 'if' logic without indentations xD
But I'll make a quick followup commit (PR) on that, which will also include a minor update in the list for the yang_modules_confd.inc file (missing a module).

troglobit:
We've had parts of confd spread out all over the tree before, and it was very difficult to keep track of origin and use.  So we gathered the pieces and put them back into src/confd/ where they belong.

Think of it like this: if confd was an external project (not living in src/) with a base set of YANG models and tools to manage it, we'd use package/confd/ to integrate it into Infix and any Infix extensions would live in board/common/.

So yes, the .inc files should live in src/confd/, next setup.sh in the scripts/ sub-directory.

ahmkar94:
Created a followup PR https://github.com/kernelkit/infix/pull/487

troglobit:
FYI: just merged main into my branch ... and I had zero (0!) conflicts?  Took me a while to realize, but in this PR, you forgot to remove the old `confd/scripts/setup.sh` :smile: 

No worries, I'll migrate my local changes of that script to the new and remove `confd/scripts/setup.sh` on my branch.

ahmkar94:
> FYI: just merged main into my branch ... and I had zero (0!) conflicts? Took me a while to realize, but in this PR, you forgot to remove the old `confd/scripts/setup.sh` 😄
> 
> No worries, I'll migrate my local changes of that script to the new and remove `confd/scripts/setup.sh` on my branch.

Ohhhh... 
Thnx :-)

PR #522, opened 2024-07-02
Bump libyang, libnetconf2, sysrepo and netopeer2

mattiaswal:
For netopeer2 we need to patch away local-users, since if it is enabled netopeer2 requires all users that should be allowed to use netconf to also be configured in ietf-netconf-server.

Also add some upstream patches.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Build related changes

### DIFF
--- a/src/confd/yang/confd.inc
+++ b/src/confd/yang/confd.inc
@@ -43,7 +43,6 @@ MODULES=(
 	"ietf-netconf@2013-09-29.yang -e writable-running -e candidate -e rollback-on-error -e validate -e startup -e url -e xpath -e confirmed-commit"
 	"ietf-truststore@2023-12-28.yang -e central-truststore-supported -e certificates"
 	"ietf-keystore@2023-12-28.yang -e central-keystore-supported -e inline-definitions-supported -e asymmetric-keys -e symmetric-keys"
-	"ietf-ssh-server@2023-12-28.yang -e local-user-auth-password -e local-user-auth-publickey"
 	"ietf-tls-server@2023-12-28.yang -e server-ident-raw-public-key -e server-ident-x509-cert"
 	"ietf-restconf@2017-01-26.yang"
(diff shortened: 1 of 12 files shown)


mattiaswal:
make legal-check: OK
tests on hw: PASS

troglobit:
Looks good, how about a fresh new changelog to kick off the new release cycle?

src/confd/yang/confd.inc
```diff
@@ -43,7 +43,6 @@ MODULES=(
 	"ietf-netconf@2013-09-29.yang -e writable-running -e candidate -e rollback-on-error -e validate -e startup -e url -e xpath -e confirmed-commit"
 	"ietf-truststore@2023-12-28.yang -e central-truststore-supported -e certificates"
 	"ietf-keystore@2023-12-28.yang -e central-keystore-supported -e inline-definitions-supported -e asymmetric-keys -e symmetric-keys"
-	"ietf-ssh-server@2023-12-28.yang -e local-user-auth-password -e local-user-auth-publickey"
 	"ietf-tls-server@2023-12-28.yang -e server-ident-raw-public-key -e server-ident-x509-cert"
```
troglobit, line 46: This is the only part I'm slightly worried about, shouldn't it match the changes to netopeer2/scripts/common.sh?

troglobit (approved):

src/confd/yang/confd.inc
mattiaswal: Could do, but not required. I think we should remove all that we do not need to change from that file. If we should add extra features it should of course remain.

troglobit: Yeah, valid point!  

Nice work on this, btw!

Issue #531, opened 2024-07-04
lower-layer-if inferred to vlanX when creating vlanX in CLI


jovatn:
### Current Behavior

Creating a vlan interface (here vlan30) via CLI sets interface type to "vlan" and fills out vlan tag-type (c-vlan) and id (30), which is helpful defaults. However, the lower-layer-if is set to itself (here vlan30), which is a bad dummy value. If executing leave without changing it, we get an error message. 
```
admin@infix-06-11-00:/> configure
admin@infix-06-11-00:/config/> edit interface vlan30
admin@infix-06-11-00:/config/interface/vlan30/> show
type vlan;
vlan {
  tag-type c-vlan;
  id 30;
  lower-layer-if vlan30;
}
admin@infix-06-11-00:/config/interface/vlan30/> leave
Error: EV ORIGIN: SHM event "done" ID 4 processing timed out.
Please check your changes, try 'diff' and 'do show interfaces'.
admin@infix-06-11-00:/config/interface/vlan30/>
```

Syslog shows the following
```
Jul  4 09:13:34 infix-06-11-00 confd[2497]: find: /run/net/3/dag/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30/vlan30: Too many levels of symbolic links
Jul  4 09:13:34 infix-06-11-00 dagger[2497]: Aborting: /run/net/3/action/init/vlan30/50-init.ip failed with exitcode 1
Jul  4 09:13:34 infix-06-11-00 dagger[2497]: Abandoned generation 3
Jul  4 09:13:34 infix-06-11-00 confd[2497]: Failed to apply interface configuration
Jul  4 09:13:34 infix-06-11-00 confd[2497]: [ERR] Oups, error detected in SR_EV_DONE 
```

And it turns out that we are stuck. Setting the lower-layer-if after a "leave" does not work. 
``` 
admin@infix-06-11-00:/config/interface/vlan30/> set vlan lower-layer-if e9
Error: EV ORIGIN: SHM event "update" ID 5 processing timed out.
[...]

PR #535, opened 2024-07-08
Initial RISC-V (riscv64) support

troglobit:
## Description

The main purpose of this PR is to add support for the [StarFive VisionFive2](https://www.starfivetech.com/en/site/boards), the first RISC-V (riscv64) architecture in Infix.  However, along the way several bugs and bad behavior that prevented this port from operating satisfactory were found and fixed.

Likely the most contentious change is that in `statd` (yanger) where instead bugging out when `ethtool --json -S` fails, we now just log a warning and proceed.  This was necessary to allow `show interfaces` to work properly.

## Other information

71IEQYaQ09L _SL1500_

Purchased at Amazon: https://www.amazon.com/dp/B0C7MV1RJ8?ref=emc_s_m_5_i_atc

## Checklist

- [X] Bugfix
  - [ ] Regression tests
- [X] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [X] Code style update (formatting, renaming)
- [ ] Build related changes


src/libsrx/src/common.h
```diff
@@ -26,6 +26,7 @@ int asprintf(char **strp, const char *fmt, ...);
 #define INFO(fmt, ...)  syslog(LOG_INFO, fmt, ##__VA_ARGS__)
 #define NOTE(fmt, ...)  syslog(LOG_NOTICE, fmt, ##__VA_ARGS__)
 #define WARN(fmt, ...)  syslog(LOG_WARNING, fmt, ##__VA_ARGS__)
+#define PANIC(fmt, ...) syslog(LOG_EMERG, fmt, ##__VA_ARGS__)
```
wkz, line 29: To me, "panic" suggests that the system will be halted/rebooted or similar, not just that a log message is emitted.

In keeping with the other log macros, maybe we should stick to the syslog level names? I.e. change this macro to `EMERG()`?

src/confd/src/ietf-interfaces.c
```diff
@@ -26,7 +26,7 @@
 #define ONOFF(boolean) boolean ? "on" : "off"
 
 #define IF_XPATH "/ietf-interfaces:interfaces/interface"
-
+#define IX_VPATH "%s/infix-interfaces:vlan"
```
wkz, line 29: Both of these are XPATHs, we don't have any VPATHs. So maybe this should be `IX_VLAN_XPATH` or something?

src/confd/src/ietf-interfaces.c
```diff
@@ -144,27 +144,33 @@ static int ifchange_cand_infer_vlan(sr_session_ctx_t *session, const char *path)
...
 		*ptr++ = '\0';
-		vid = strtol(ptr, NULL, 10);
+		vid    = strtol(ptr, NULL, 10);
+		lower  = ifname;
 	} else if (!fnmatch("vlan+([0-9])", ifname, FNM_EXTMATCH)) {
 		if (sscanf(ifname, "vlan%d", &vid) != 1)
 		    goto out_free_ifname;
+
+		/* Avoid setting lower-layer-if to vlanN */
+		lower = NULL;
```
wkz, line 154: What would be the harm of setting `lower = "br0";` here? In the common case where it exists, we save a few keystrokes; if not, then the user has to fill it in for themselves.

board/riscv64/visionfive2/rootfs/etc/rauc/system.conf
```diff
@@ -0,0 +1,18 @@
...
+mountprefix=/var/lib/rauc/mnt
+bundle-formats=-plain
+max-bundle-download-size=1073741824
+
+[keyring]
+directory=/etc/rauc/keys
+
+[slot.rootfs.0]
+device=/dev/mmcblk1p3
+bootname=primary
```
wkz, line 14: If you change the partition named `rootfs` in `genimage.cfg` to `primary`, can't you use the same device path as on `aarch64` here?

wkz (approved):
Awesome work! :100: 

Now I _really_ want one of these boards :smile:

Only found some stylistic things.

src/libsrx/src/common.h
troglobit: You caught that green goblin, like catnip to you, those goblins :smirk_cat: 

Will fix, was debating with myself long and hard between `EMERG()` and `PANIC()`, but I agree with you.  We should save the latter for cases when we need to actually panic (and possibly reboot using `watchdogd`).

board/riscv64/visionfive2/rootfs/etc/rauc/system.conf
troglobit: Reworking the entire thing right now. Only reason for keeping it named `rootfs` as above was to have the image be loaded from the default U-Boot that comes preloaded on the QSPI flash.

Will make a follow-up commit to address this, thanks!

src/confd/src/ietf-interfaces.c
troglobit: Yeah I took the conservative approach from #531.  We should probably pull in @jovatn into this discussion, to me it's fine either way as long as we don't set `lower` to the same as `ifname` (the bug).

src/confd/src/ietf-interfaces.c
troglobit: Agreed, fixing!

src/confd/src/ietf-interfaces.c
wkz: I'll pick his brain on this AFK, thanks!

Now that I think about it, shouldn't there also be a `must` expression in the model that forbids setting the lower to itself? Otherwise we are still vulnerable to the user making the same mistake, right?

mattiaswal (approved):

troglobit:
> Now I _really_ want one of these boards :smile:

It's actually a quite decent board! I have the mechanic enclosure with cooling tower, which was a great deal 👍

src/confd/src/ietf-interfaces.c
troglobit: Skipping.  AFK discussion decided current fix is the best one for now.

mattiaswal (approved):

Issue #158, opened 2023-10-06
Properly wipe /cfg partition on factory reset


troglobit:
I.e., in `board/common/rootfs/lib/infix/mnt`, write zeroes to the partition to be mounted on `/cfg`

PR #551, opened 2024-08-11
Update NanoPi R2S

### LINKED ISSUE #274: NanoPi R2S: traffic LEDs on eth1 (LAN) not working
While porting to Infix I'm sure I had the LEDs on both ports working fine at some point.  Went from Linux 5.16, and while trying out various WiFi USB plugs, I went as high as 6.7.1 before landing on 6.5.11, which Infix is currently is on.  So current `linux_defconfig` works on that newer kernels, but we have some patches that don't apply, so no way of testing before relocating them to a subdirectory.

From discussion in https://github.com/kernelkit/infix/pull/271#issuecomment-1905381201

### LINKED ISSUE #275: aarch64: reorg. documentation a bit
The current set of READMEs in board/aarch64/ are a bit of an incoherent mess.

 - Move board/aarch64/README.md to a subdirectory for the SparX-5i
 - Add missing docs for EspressoBIN, R2S (console port pinout, sdcard.img, and LED issue(s), and CN9130-CRB
 - Detail paths for how to add new boards, what qualifies for `aarch64_defconfig`, etc.

### LINKED ISSUE #542: login: warning missing pam_lastlog.so
```
login[2819]: PAM unable to dlopen(/lib/security/pam_lastlog.so): /lib/security/pam_lastlog.so: cannot open shared object file: No such file or directory
```

troglobit:
## Description

This PR holds massive updates of the Infix support to the NanoPi R2S, initially added in #271.  From the ChangeLog:

- Update Linux kernel to v6.10.3 and sync defconfig with aarch64
- Workaround `reboot` command "hang" on NanoPi R2S (failure to reboot) by replacing the Rockchip watchdog driver with "softdog"
- Update U-Boot to v2024.07, enable secure boot loading of images
- Rename interfaces to LAN + WAN to match case and LEDs
- Rename images to `infix-r2s$ver.ext`, not same as other aarch64
- Change rootfs to squashfs for enhanced security
- Add RAUC support to simplify device maintenance/upgrade
- Add support for saving unique interface MAC addresses in U-Boot
- Add support for system LEDs, see product's README
- Add support for reset button from Linux, issues `reboot` atm.
- Add static `factory-config` as an example
- Add LED indication on factory reset, *all* LEDs available in Linux `/sys/class/leds` are turned on while clearing writable partitions
- Issue #158: enhance security of factory reset.  All file content is now overwritten x3, the last time with zeroes, then removed.  Example, on the NanoPi R2S this process takes ~30 seconds, but may take longer in setups with bigger configurations, e.g., containers
- Fix #274: add missing link/traffic LEDs on NanoPi R2S LAN port
- Fix #275: add missing documentation for the R2S (EspressoBin coming later)
- Fix #542: warning message from `login`, cannot find `pam_lastlog.so`

## Checklist

- [X] Bugfix
  - [ ] Regression tests
- [X] Feature
  - [ ] Regression tests added?
  - [X] Documentation added?
- [X] Build related changes


doc/ChangeLog.md
```diff
@@ -9,15 +9,36 @@ All notable changes to the project are documented in this file.
 
 ### Changes
 - Initial RISC-V (riscv64) support: StarFive VisionFive2
+- Massive updates to the NanoPi R2S:
+  - Update Linux kernel to v6.10.3 and sync defconfig with aarch64
```
mattiaswal, line 13: I do not really like this, i think all platform should use the same kernel (if possible). To ease maintainability for potental bugfixes etc. To get But that is just my principles, please challenge it.

troglobit: Interesting point of discussion.  I think of it more as a long-term goal to work towards rather than an absolute.  In particular to keep differences between Infix-based boards to a minimum, but unlike previous places of employment I'm not as conservative now.

I see both the NanoPi and the VisionFive boards as references for Infix, to help lower the barrier of entry, ease adoption, and show off the operating system's capabilities.  In that sense some boards will have features that it doesn't share with others, even though they share the same base architecture.  E.g., with/without: switchcore, wifi, cellular, etc.

It's also a question of the number of users on a board, paying customers vs freebie users.  I don't think we should spend too much time on maintaining and testing support for all boards in a common kernel, for instance.  I tried that with the EspressoBIN and MacchiatoBIN boards, but the pace at which our development progressed on the `aarch64_defconfig` it was impossible to keep them working -- same with the CRB board.  I've actually planned to break out more boards from the standard aarch64 build to be able to maintain them more easily, because man power is another component we need to consider here.

So, that's where I'm coming from.  Love to hear what @wkz has to say on this as well.

mattiaswal: What i mean is if we use the same kernel for the "platform" we could for example build a small test system with R2S (or proffered,  something with 2 more ports) and actually gain from it.

And if a (non-maintained) board breaks after an upgrade, so be it, open a bug report. But at least it has the same patch level as all other supported boards and we can draw conclusions about behavior.

troglobit: OK, we've had a talk about this AFK now too.  Can we agree on merging this for now, and postpone the merge to the `aarch64_defconfig` in the next PR, @lazzer + @wkz ?

Then I could base my planned refresh of the EspressoBIN support on this and coax it all together in the same build for that PR.  I'd like to get on that as soon as possible, even this coming weekend :nerd_face:

mattiaswal: Thats fine for me.

mattiaswal (approved):

wkz (approved):
I vote for merging this for now. I really hope we can get this to a state where it becomes a fully supported Tier 1 board soon though.

Side note: when that happens, I think we should move lots of the excellent docs in this PR, regarding LEDs, factory reset, etc., to the general documentation :+1:

Issue #585, opened 2024-08-22
Failure in bridge configuration lock sysrepo


mattiaswal:
Very easy to reproduce. I use qemu with 6 interfaces

<pre>
admin@infix-00-00-00:/> copy fa ru
admin@infix-00-00-00:/> configure 
admin@infix-00-00-00:/config/> set interface br1
admin@infix-00-00-00:/config/> set interface e3 bridge-port bridge br1
admin@infix-00-00-00:/config/> set interface br1 bridge vlans vlan 30 multicast-filter 224.2.2.2 ports e3
admin@infix-00-00-00:/config/> leave 
Error: EV ORIGIN: SHM event "done" ID 2 processing timed out.
Please check your changes, try 'diff' and 'do show interfaces'.
admin@infix-00-00-00:/config/> 
admin@infix-00-00-00:/config/> abort
admin@infix-00-00-00:/> reboot
Error: EV ORIGIN: SHM event "rpc" ID 1 processing timed out.
Error: failed sending RPC /ietf-system:system-restart: User callback failed
admin@infix-00-00-00:/> 
</pre>

This config is wrong (e3 does not exist in vlan 30) and should be blocked by a must expression, but still this feel fishy.

mattiaswal:
Failed even with correct VLAN setting (e3 untagged).
If i do it in small steps it succeed. Configure one thing and leave.

troglobit:
Reproduced. Have possible sysrepo fix in my tree.

Issue #574, opened 2024-08-22
Run 'make legal-info' more often, not only at release builds


troglobit:
The v24.06.0-rc1 build failed due to `make legal-info` not working.

We had silly mistakes like:

 - LICENCE vs LICENSE
 - LICENCE vs COPYING
 - Missing hash update in Buildroot package bump

We should probably add legal-info to Bob and run on all PRs as well.

Issue #541, opened 2024-07-14
Frr logs to /var/log/frr without rotation


troglobit:
Set up Frr with syslog, format rfc5424:

- https://docs.frrouting.org/en/latest/basic.html#clicmd-log-syslog-LEVEL
- https://docs.frrouting.org/en/latest/extlog.html#clicmd-destination-syslog-supports-rfc5424

This will let it log structured logs directly to `/dev/log`, without passing the non-functional GLIBC `syslog()` API.  Allowing `syslogd` to both filter, forward, and handle log rotation.

> This works because the Infix syslog daemon is [sysklogd](https://github.com/troglobit/sysklogd), which fully supports RFC5424.

PR #588, opened 2024-08-26
Fix blocking sysrepo on confd internal error

troglobit:
## Description

This PR fixes a serious blocking issue in sysrepo caused by `sysrepo-plugind` not handling errors in DONE callbacks properly.  A few other fixes and improvements are also included:

Fixes:

 - #585 
 - #541 
 - #574 

## Checklist

- [X] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Build related changes


mattiaswal (approved):
:+1:  My only opinion is that enable debug in OSPF may impact performance/stability on OSPF, but we test with it enabled.

troglobit:
> 👍 My only opinion is that enable debug in OSPF may impact performance/stability on OSPF, but we test with it enabled.

Agreed, let's leave it on to gauge the impact.  Like we talked about AFK it would be great to have a knob for debug in YANG, I'll make sure to add a separate task for that before merging this.

Still tracking down the issues I see with the tests.  Seems to be a `syslogd` issue so far.

troglobit:
> > 👍 My only opinion is that enable debug in OSPF may impact performance/stability on OSPF, but we test with it enabled.
> 
> Agreed, let's leave it on to gauge the impact. Like we talked about AFK it would be great to have a knob for debug in YANG, I'll make sure to add a separate task for that before merging this.
> 
> Still tracking down the issues I see with the tests. Seems to be a `syslogd` issue so far.

Further testing gives that `debug` gives *quite a lot* more logs than I initially could see.  So I'm reverting back to `informational` -- keeping the `debug ospf *` though, making it easier to enable debug logs if needed.

test/env
```diff
@@ -199,6 +201,7 @@ if [ "$containerize" ]; then
 	 --env PROMPT_DIRTRIM="$ixdir" \
 	 --env PS1="$(build_ps1)" \
 	 --expose 9001-9010 --publish-all \
+	 -v "$HOME/.ash_history":/root/.ash_history \
```
mattiaswal, line 204: Is it really good to get $HOME/.ash_history, that may contain the history from the host system? Isn't something infix specific better: $HOME/.infix/ash_history

troglobit: Hmm, good point.  Iirc we have some other file/dir that we save to `$HOME` as well ... maybe @wkz remembers?

wkz: The python venv is stored in `~/.infix-test-venv`, maybe that's the one you're thinking of?

IIRC, it takes a while to compose it, so I decided to cache it there to save some time. But it might be better to move that as well :man_shrugging:

troglobit: Ah, of course, thanks!  I'll look into moving that as well, in under a common `~/.infix/`

troglobit:
@mattiaswal + @wkz there, done.  Unless you have any further comments, I'll merge this as soon as the build + test finishes.  Thanks for the comments so far!

PR #590, opened 2024-08-29
Misc fixes

mattiaswal:
* Fix test for syslog (prefix some settings with infix-syslog)
* Improve must expression for multicast-filters in infix-if-bridge


troglobit (approved):
No notes :fire:

Issue #597, opened 2024-09-03
Add support for minimal defconfigs


troglobit:
In an effort to streamline what are the *core Infix* requirements, as well speed up builds both in CI and on developer machines, we have arrived at a need for `$ARCH_minimal_defconfig` files.

### Tasks

 - [x] Replace the default builds in CI with minimal defconfigs
 - [x] Allow checkbox or ~~a CI keyword, similar to [[skip ci]](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/skipping-workflow-runs)~~ ... or pull request labels(!), to trigger default defconfigs
 - [x] Adapt Infamy to cope with restrictions of minimal builds, e.g., no containers or RESTCONF
 - [x] If default defconfigs are also selected for a job, use those images for Infamy
 - [x] Include Buildroot `.config` in image, similar to `/boot/config.gz`, to know exactly what is running
 - [x] Modify `/etc/os-release` for minimal builds
 - [x] How to identify, different hostname (`ix`) and bootup msg?

### Definition of minimal defconfig

 - No container support => no need to build Go compiler for podman
 - Smallest possible *CONF => NETCONF today
 - No extra/future packages, not even Tetris

> *How low can you go?*

troglobit:
Turns out the best (?) or most intuitive way to control the flow for pull requests is to set a label when creating the PR, or afterwards (but before the next push to the branch).  

I've created the `ci:main` label to force branch builds to use the standard defconfigs.  This can also be triggered when starting jobs manually (workflow dispatch) from the Actions menu.  A dedicated checkbutton 'minimal' has been added, checked true by default.

PR #599, opened 2024-09-03
Minimal defconfig

troglobit:
## Description

This PR adds support for a set of minimal defconfigs for Infix tier one targets, including Infamy adaptations, some identifying markers, and new mechanisms for controlling the build of pull requests.

For the full specification, see:
- #597

As a result, the build times have been slashed in half (!), even on our latest build server.  Developer laptops should see a similar speedup.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [X] Build related changes


test/test.mk
```diff
@@ -16,11 +16,15 @@ mode-host   := -t $(or $(TOPOLOGY),/etc/infamy.dot)
...
 mode        := $(mode-$(TEST_MODE))
 
-
 binaries-$(ARCH) := $(addprefix $(IMAGE)-$(ARCH),.img -disk.img .pkg)
 binaries-x86_64  += OVMF.fd
 binaries := $(foreach bin,$(binaries-$(ARCH)),-f $(BINARIES_DIR)/$(bin))
 
+# Common transport override for minimal defconfigs
+ifneq ($(BR2_PACKAGE_ROUSETTE),y)
+export INFAMY_TRANSPORT := netconf
```
wkz, line 25:
```suggestion
export INFAMY_ARGS := --transport netconf
```
I think we can already pass infamy arguments via this variable. This is how `test/env` passes the physical topology path to the tests, for example.

test/case/meta/wait.py
```diff
@@ -19,6 +22,26 @@ def ll6ping(node):
...
+            return netconf.netconf_syn(neigh)
+        elif env.transport == "restconf":
+            return restconf.restconf_reachable(neigh, password)
+        else:
+            raise Exception(f"Unsupported transport {env.transport}!")
+
+    netconf_reachable = netconf.netconf_syn(neigh)
+    restconf_reachable = restconf.restconf_reachable(neigh, password)
+
+    return netconf_reachable and restconf_reachable
```
wkz: If this is moved to `util.py`, couldn't we reuse it at the end of `wait_boot()` to get rid of the code duplication?

I.e. replace that with

```python
until(lambda: is_reachable(...), attempts=300)
```

test/infamy/transport.py
```diff
@@ -54,6 +54,23 @@ def get_mgmt_iface(self):
...
+
+        :param model_name: Name of the YANG model to check.
+        :param xpath: Optional XPath to refine search to a specific property.
+        :return: True if the model is found, False otherwise.
+        """
+        # If xpath is provided, extract the module name from it
+        if xpath:
+            mod = [seg.split(":")[0] for seg in xpath.split("/") if ":" in seg]
+            if model_name not in mod:
+                return False
```
wkz, line 69: I can't wrap my head around how this is supposed to be used - and it seems there are no users in this PR at least. Do we need it?

wkz (changes requested):
I did not expect you to pull this off this fast - excellent! :clap:

test/test.mk
troglobit: Sure, works with an extra `--env INFAMY_ARGS="$INFAMY_ARGS"` to docker wrapper.

I tried almost this first, but using `INFAMY_EXTRA_ARGS` which blew` up on me and I never got working with `make test-sh`.

test/case/meta/wait.py
troglobit: Put it as future work in the commit message, but I can have a look.

test/infamy/transport.py
troglobit: It was for checking for features in the model, not used atm., but since we don't look for revision I though this could come in handy.

troglobit:
> `3333 execd running [---2345----] Container job runner`
> 
> is this by design?

Ah yes, `execd`, the general queue executor that became more or less hard coded to only work for containers in Infix ... it should probably not be built or run on minimal builds.  I can have a look and see what I can do about cleaning up that mess a bit.  I should probably add a ticket to refactor the job runner to use [ts (task spooler)](http://viric.name/soft/ts/) instead.

test/infamy/transport.py
troglobit: Removed in my local queue.

troglobit:
> > `3333 execd running [---2345----] Container job runner`
> > is this by design?
> 
> Ah yes, `execd`, the general queue executor that became more or less hard coded to only work for containers in Infix ... it should probably not be built or run on minimal builds. I can have a look and see what I can do about cleaning up that mess a bit. I should probably add a ticket to refactor the job runner to use [ts (task spooler)](http://viric.name/soft/ts/) instead.

Fixed.  Thank you for spotting this one!

troglobit:
@mattiaswal + @wkz all comments and outstanding issues have now been fixed.
Please check again and/or ... trust me bro! :sunglasses:

wkz (approved):
Fuiyoh!

troglobit:
> Fuiyoh!

Uncle Troglo represent! 😂

troglobit:
OK, merging, @mattiaswal approved AFK.

PR #615, opened 2024-09-10
Fix eui 64 based ipv6 addresses

rical:
## Description
 
This PR fixes the EUI-64 based IPv6 autoconf address on bridges and updates the IGMP basic test.                                                                                                                                            
                                                                                                                                                                                                                                            
The EUI-64 address is fixed by explicitly assigning the base mac address to the bridge when creating it. Strangely, this messes with some timing in the bridge, causing the first multicast querier message sent to the bridge from user-space to be lost. This is caused by the bridge using the "NOOP" scheduler (noop_enqueue()) for a short period of time while the interface is starting up. To avoid the test-case failing when this happens we introduce a retry that will wait for the second query message. This is a bit strange in the case of a software device such as a bridge, i.e. it might be a bug. However, loosing the first query isn't strange in the case of a normal HW interface. That's how we justify the retry loop.

### DIFF
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -357,6 +357,10 @@ future releases.
 - Fix #349: minor changes to `bridge-port` settings, like setting `pvid`
   when you forget it, did not take without a reboot
 - Fix #353: impossible to remove bridge port with `no bridge-port`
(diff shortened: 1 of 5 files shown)


troglobit (approved):
Great stuff, thank you so much for diving down this rabbit hole! :1st_place_medal:

PR #620, opened 2024-09-12
Misc bugfixes

troglobit:
## Description

Misc. fixes to annoying regressions and missing functionality.

## Checklist

- [X] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Build related changes

### DIFF
--- a/doc/container.md
+++ b/doc/container.md
@@ -222,9 +222,8 @@ archive, which helps greatly with container upgrades (see below):
 Upgrading a Container Image
 ---------------------------
 
-> **Note:** the default writable layer is lost when upgrading the image
-> Use named volumes for directories with writable content you wish to
-> keep over an upgrade.
+> **Note:** the default writable layer is lost when upgrading the image.
+> Use named volumes for content that you want to persist across upgrades.
 
 All container configurations are locked to the image hash at the time of
 first download, not just ones that use an `:edge` or `:latest` tag.  An
--- a/src/statd/python/yanger/yanger.py
+++ b/src/statd/python/yanger/yanger.py
@@ -528,7 +528,9 @@ def add_container(containers):
 
 
 def get_brport_multicast(ifname):
-    data = run_json_cmd(['mctl', 'show', 'igmp', 'json'], "igmp-status.json")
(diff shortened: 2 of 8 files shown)


src/statd/python/yanger/yanger.py
```diff
@@ -102,7 +102,7 @@ def insert(obj, *path_and_value):
     curr[path[-1]] = value
 
 
-def run_cmd(cmd, testfile, default=None):
+def run_cmd(cmd, testfile, default=None, quiet=False):
```
wkz, line 105: For people familiar with Python's `subprocess.run()`, I think `check` is more familiar.

```suggestion
def run_cmd(cmd, testfile, default=None, check=True):
```

Then we can change `s/subprocess.check_output(...)/subprocess.run(..., check=check).output` below.

That way, a non-zero exitcode won't raise an exception, so there's no need for the `if not quiet` guard below.

doc/container.md
```diff
@@ -222,9 +222,9 @@ archive, which helps greatly with container upgrades (see below):
 Upgrading a Container Image
 ---------------------------
 
-> **Note:** the default writable layer is lost when upgrading the image
-> Use named volumes for directories with writable content you wish to
-> keep over an upgrade.
+> **Note:** the default writable layer is lost when upgrading the image.
+> Use named volumes for directories with writable content that you wish
+> to keep over an upgrade.
```
wkz, line 227:
```suggestion
> Use named volumes for directories with writable content that you wish
> to keep over an upgrade.
```
```suggestion
> Use named volumes for content that you want to persist across upgrades.
```

Something about "keep over an upgrade" doesn't feel quite right, but this is just my opinion.

wkz (approved):
Great work tracking this down! :heart: 

Just found a couple of stylistic things, do with them what you will.

src/statd/python/yanger/yanger.py
troglobit: Ah, of course, great idea thanks!

PR #651, opened 2024-09-25
Statd: improve speed of interface data 

rical:
## Description
This PR removes the statd interface netlink code and improve speed of operational interface data.
    
This is a major redesign of statd interface handling. The goal here is to reduce the time it takes for statd to produce operational data for all interfaces on larger iron. Which manifested itself most notably in the CLI command "show interfaces".
    
This PR removes the netlink interface subscriptions done by the statd c code. Instead statd now subscribes to the ietf-interfaces base path which means it will be invoked for all interfaces. If the user (sysrepo) has specified an explicit interface name such as eth0, it will be extracted by statd and passed down to Yanger.
    
This PR also modifies Yanger to fit this new modus. No explicitly passed interface name now means all interfaces.
    
 ## Side note
The main culprit of the slow operational interface data is the time it takes to bootstrap python. Which prior to this PR was done for each interface. On the HW system I'm testing this PR on, runtime is reduced by a factor of 10.

### DIFF
--- a/src/statd/python/yanger/yanger.py
+++ b/src/statd/python/yanger/yanger.py
@@ -837,6 +837,26 @@ def add_vlans_to_bridge(brname, iface_out, mc_status):
 
     insert(iface_out, "infix-interfaces:bridge", "vlans", "vlan", vlans)
 
+def get_iface_data(ifname):
+    iface_out = {}
+
+    add_ip_link(ifname, iface_out)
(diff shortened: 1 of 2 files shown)


troglobit (approved):
Amazing work on this! :fire: :clinking_glasses: 

Must say I also, very much, appreciate the detailed analysis in the report and summary of changes here in the PR. :1st_place_medal:

Issue #658, opened 2024-09-26
confd: fails removing veth pair after a multple reconfigurations


ahmkar94:
### Current Behavior

The NETCONF request provided creates a bridge `br-X` with the port `ethX` (mapped to a specific port on the target). In the example below, the test initializes the environment and attaches the target, then configures the bridge and its associated interface:

```
with infamy.Test() as test:
    with test.step("Initialize"):
        env  = infamy.Env()
        target = env.attach("target", "mgmt")

        _, eth_X = env.ltop.xlate("target", "ethX")
        br_X = "br-X"

    with test.step("Configure bridge brX and associated interfaces"):
        target.put_config_dict("ietf-interfaces", {
        "interfaces": {
            "interface": [
                {
                    "name": br_X,
                    "type": "infix-if-type:bridge",
                    "enabled": True
                },
                {
                    "name": eth_X,
                    "type": "infix-if-type:ethernet",
                    "enabled": True,
                    "infix-interfaces:bridge-port": {
                        "bridge": br_X
                    }
                }
            ]
        }
    })
```
By looking at the logs on the target we see that `confd` crashes:
```
Sep 26 16:13:21 test-00-01-00 dagger[3255]: Aborting: /run/net/2/action/init/e3/10-ethtool-aneg.sh failed with exitcode 75
Sep 26 16:13:21 test-00-01-00 dagger[3255]: Abandoned generation 2
Sep 26 16:13:21 test-00-01-00 confd[3255]: Failed to apply interface configuration
Sep 26 16:13:21 test-00-01-00 confd[3255]: Oups, error detected in SR_EV_DONE
Sep 26 16:13:21 test-00-01-00 confd[3255]: failed sr_subscription_process_events(), ret:7
Sep 26 16:13:21 test-00-01-00 finit[1]: Stopping netopeer[3771], sending SIGTERM ...
Sep 26 16:13:21 test-00-01-00 finit[1]: Stopping statd[3716], sending SIGTERM ...
Sep 26 16:13:21 test-00-01-00 finit[1]: Stopping rousette[3801], sending SIGTERM ...
Sep 26 16:13:21 test-00-01-00 finit[1]: Service confd[3255] died, restarting in 2000 msec (1/10)
Sep 26 16:13:21 test-00-01-00 finit[1]: Starting confd[4462]
```

[...]

PR #661, opened 2024-09-27
Fix test issues after adding veth-delete test

troglobit:
## Description

 - Fixes a regression introduced on `main` while fixing issue #658
 - Check for `control-plane-protocol` in `startup-config` before migrating and causing empty `startup-config`
 - Return `None` on 404 return for fetching a given operational XPath in RESTCONF test backend
 - Fix `tcpdump` permissions for `admin` user in CLI

## Checklist

- [x] Bugfix
  - [x] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Build related changes

### DIFF
--- a/test/infamy/restconf.py
+++ b/test/infamy/restconf.py
@@ -207,7 +207,13 @@ class Device(Transport):
             dspath = f"{dspath}/{path}"
 
         url = f"{self.restconf_url}{dspath}"
-        return self._get_raw(url, parse)
+        try:
+            return self._get_raw(url, parse)
+        except requests.exceptions.HTTPError as e:
+            if e.response.status_code == 404:
+                return None
+            else:
+                raise e
 
     def get_running(self, path=None):
         """Wrapper function to get running datastore"""
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -83,6 +83,8 @@ documentation for details.
   interface changes (tries to delete both ends of the pair)
 - Spellcheck path to `/var/lib/containers` when unpacking OCI archives
   on container upgrade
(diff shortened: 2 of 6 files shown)


test/infamy/restconf.py
```diff
@@ -207,7 +207,13 @@ def get_datastore(self, datastore="operational", path="", parse=True):
             dspath = f"{dspath}/{path}"
 
         url = f"{self.restconf_url}{dspath}"
-        return self._get_raw(url, parse)
+        try:
+            return self._get_raw(url, parse)
+        except Exception as e:
```
wkz, line 212:
```suggestion
        except requests.exceptions.HTTPError as e:
```
If something else goes wrong, I think we risk triggering an `AttributeError` when trying to access `e.response` below, loosing the original exception information.

wkz (approved):
Possible improvement. Probably unlikely to trigger another exception, then again, it will be that much hard to find in that case :man_shrugging:

test/infamy/restconf.py
troglobit: True, thanks!

I have more fixes coming. QoS has stopped working as well, for some reason.

test/infamy/restconf.py
```diff
@@ -207,7 +207,13 @@ def get_datastore(self, datastore="operational", path="", parse=True):
             dspath = f"{dspath}/{path}"
 
         url = f"{self.restconf_url}{dspath}"
-        return self._get_raw(url, parse)
```
mattiaswal, line 210: Not a fan of this. This will hide errors. You only get 404 when you do something wrong

mattiaswal (approved):
Go go go! :boom:

PR #665, opened 2024-09-30
Do not get internal interfaces when getting all interfaces

mattiaswal:
This is a regression from 2024.08

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Build related changes

### DIFF
--- a/src/statd/python/yanger/yanger.py
+++ b/src/statd/python/yanger/yanger.py
@@ -156,6 +156,8 @@ def iface_is_dsa(iface_in):
         return False
     return True
 
+def iface_is_internal(iface_in):
+    return iface_in.get("group") == "internal"
 
 def get_vpd_vendor_extensions(data):
     vendor_extensions = []
@@ -587,7 +589,6 @@ def get_brport_multicast(ifname):
 
     return multicast
 
-
 def add_ip_link(ifname, iface_out):
     """Fetch interface link information from kernel"""
     data = run_json_cmd(['ip', '-s', '-d', '-j', 'link', 'show', 'dev', ifname],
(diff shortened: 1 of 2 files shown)


troglobit:
Unfortunately this will not make it for the v24.09 release.

src/statd/python/yanger/yanger.py
```diff
@@ -156,6 +156,13 @@ def iface_is_dsa(iface_in):
         return False
     return True
 
+def iface_is_internal(iface_in):
+    if "group" not in iface_in:
+        return False
+
+    if iface_in["group"] != "internal":
+        return False
+    return True
```
wkz, line 165:
```suggestion
    return iface_in.get("group") == "internal"
```
`.get()` on dictionaries is pretty handy in these situations.

src/statd/python/yanger/yanger.py
```diff
@@ -587,7 +594,6 @@ def get_brport_multicast(ifname):
 
     return multicast
 
-
```
wkz, line 590: Isn't double-newlines some PEP8 thing?

wkz (approved):
Just some stylistic comments. Do with them what you will :smile:

rical:
Right, I missed that. I will have a look during the day.

src/statd/python/yanger/yanger.py
troglobit: Indeed it is, between methods/classes/etc. outside of a class, iirc.

rical:
I was already thinking about redesigning how Yanger fetches and iterates ip interface data. With this redesign this became even easier to solve. It also allowed me to cleanup the unit tests and add a test for the hiding of internal interfaces.

Have a look, it's in PR: #674 (supersedes this PR)

mattiaswal:
Obsoleted by #674

Issue #390, opened 2024-04-10
Upgrade netconf_client for tests and create a new docker image


mattiaswal:
This since they now have added get-data RPC which we have patched on our own.
infamy/netconf.py:
<pre>
 def _get_data(self, xpath,as_xml=False):
        """Local member wrapper for netconf-client <get-data> RPC"""
        # pylint: disable=protected-access
        (raw, ele) = self.ncc._send_rpc(self._ncc_get_data_rpc(filter=xpath))
        data = NccGetDataReply(raw, ele)
        if(as_xml):
            return data
</pre>
From netconf_client:
<pre>
add support for the < get-data > command from RFC 8526
</pre>

troglobit:
Ah great, finally!

PR #683, opened 2024-10-03
Upgrade NETCONF Client from 2.2.0 to 3.1.1

axkar:
Refactor NETCONF functions (get_data) to accommodate the upgrade
in the netconf_client library within the Docker container image.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Build related changes

### DIFF
--- a/test/.env
+++ b/test/.env
@@ -2,7 +2,7 @@
 # shellcheck disable=SC2034,SC2154
 
 # Current container image
-INFIX_TEST=ghcr.io/kernelkit/infix-test:1.9
+INFIX_TEST=ghcr.io/kernelkit/infix-test:2.0
 
 ixdir=$(readlink -f "$testdir/..")
 logdir=$(readlink -f "$testdir/.log")
--- a/test/infamy/netconf.py
+++ b/test/infamy/netconf.py
@@ -204,25 +204,39 @@ class Device(Transport):
         return list(filter(lambda m: m["name"] in modnames,
                            self.modules.values()))
 
+    def _build_xpath_filter(self, xpath, get_data_xpath=False):
+        """Helper function to build the XPath filter with the necessary xmlns."""
+        if xpath:
+            xmlns = " ".join([f"xmlns:{m['name']}=\"{m['namespace']}\""
+                                for m in self._modules_in_xpath(xpath)])
+            if get_data_xpath is True:
+                return f'<xpath-filter {xmlns}>{xpath}</xpath-filter>'
+            else:
+                return f"<filter type=\"xpath\" select=\"{xpath}\" {xmlns} />"
+        return None
+    
+    def _parse_response(self, response, parse):
+        """Helper function to handle XML response parsing."""
+        if not response:
+            return None
+
+        if not parse:
+            return response.raw_reply
+
+        if len(response.data_ele) == 0:
+            return None
+
+        cfg = lxml.etree.tostring(response.data_ele[0])
+        parsed_data = self.ly.parse_data_mem(cfg, "xml", parse_only=True)
+
+        return parsed_data
+
     def _ncc_make_rpc(self, guts, msg_id=None):
         if not msg_id:
             msg_id = uuid.uuid4()
 
         return '<rpc message-id="{id}" xmlns="urn:ietf:params:xml:ns:netconf:base:1.0">{guts}</rpc>' \
             .format(guts=guts, id=msg_id).encode("utf-8")
-
-    def _ncc_get_data_rpc(self, datastore="operational", filter=None, msg_id=None):
-        pieces = []
-        pieces.append('<get-data xmlns="urn:ietf:params:xml:ns:yang:ietf-netconf-nmda">')
-        pieces.append(f'<datastore xmlns:ds="urn:ietf:params:xml:ns:yang:ietf-datastores">'
-                      f'ds:{datastore}'
-                      f'</datastore>')
-        if filter:
-            xmlns = " ".join([f"xmlns:{m['name']}=\"{m['namespace']}\""
-                              for m in self._modules_in_xpath(filter)])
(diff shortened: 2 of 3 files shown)


ahmkar94:
It is probably easier to ask and comment here in the PR than in the issue #390 
@troglobit  @mattiaswal I have a question regarding `get_config` in netconf.py.

I think we should not leverage `_get` for that since it is used to fetch both `operational` and `running` (all config).
`ncc.get_config` allows for specifying `source` (datastore), as far as I see in netconf_client.

see:
https://github.com/ADTRAN/netconf_client/blob/main/netconf_client/ncclient.py
https://github.com/ADTRAN/netconf_client/blob/main/netconf_client/rpc.py

troglobit:
@ahmkar94 not sure I follow?  

The method infamy/netconf.py:_get() is (and should!) only used locally in the Class by get() and get_config().  You are free to refactor infamy/netconf.py any way you see fit to reduce the amount of extra code we have to make future upgrades of ncclient easier :smiley: :hammer_and_wrench:

test/.env
```diff
@@ -2,7 +2,7 @@
 # shellcheck disable=SC2034,SC2154
 
 # Current container image
-INFIX_TEST=ghcr.io/kernelkit/infix-test:1.9
+INFIX_TEST=ghcr.io/kernelkit/infix-test:2.0
```
troglobit, line 5: This was a good idea!  If it had been me, I'd probably have bumped it to 1.10, but since we're upgrading a major component in the image, we should probably also bump the major number of the image.

Unless @mattiaswal and @wkz disagree, I think we should update the test/docker/README.md with this recommendation for the future.

test/infamy/netconf.py
```diff
@@ -285,16 +257,28 @@ def get_dict(self, xpath):
...
 
-        return data.print_dict()
+        if not parse:
+            return response.raw_reply
+        
+        if len(response.data_ele) == 0:
+            return None
+        
+        cfg = lxml.etree.tostring(response.data_ele[0])
+        parsed_data = self.ly.parse_data_mem(cfg, "xml", parse_only=True)
```
troglobit, line 279: You removed the `pylint` comment, does that mean pylint no longer complains about this construct?

test/infamy/netconf.py
```diff
@@ -435,5 +419,12 @@ def get_current_time_with_offset(self):
...
+        parsed_data = lxml.etree.fromstring(data)
+
         xpath = './/{urn:ietf:params:xml:ns:yang:ietf-system}current-datetime'
-        return data.data_ele.find(xpath).text
+        current_datetime = parsed_data.find(xpath)
+
+        if current_datetime is not None:
+            return current_datetime.text
+        else:
+            raise ValueError("current-datetime element not found in the response")
```
troglobit, line 430: It's a good practice to ensure all files have a final newline.  (It's been removed in this patch.)

troglobit (changes requested):
Some minor things only.

I can checkout your branch and push a new infamy docker image.

ahmkar94:
> @ahmkar94 not sure I follow?
> 
> The method infamy/netconf.py:_get() is (and should!) only used locally in the Class by get() and get_config(). You are free to refactor infamy/netconf.py any way you see fit to reduce the amount of extra code we have to make future upgrades of ncclient easier 😃 🛠️

I meant that as you said _get is used by both get() and get_config(), but _get does not allow for the `source` parameter to be passed when get_config is used as a getter function.This way get_config will always fetche the default (running configuration) when calling it (as far as I understand) :)
```
def get_config(self, source="running", filter=None, with_defaults=None):
```

ahmkar94:
But I can make an additional commit for that anyways and we see then :)

test/infamy/netconf.py
```diff
@@ -285,16 +257,28 @@ def get_dict(self, xpath):
...
+        if not response:
             return None
 
-        return data.print_dict()
+        if not parse:
+            return response.raw_reply
+        
+        if len(response.data_ele) == 0:
+            return None
+
```
troglobit, line 277: Just noticed that there are extra whitespace on the two blank lines surrounding this if-statement.  Another thing pylint will complain about.

troglobit:
> I can checkout your branch and push a new infamy docker image.

Container image updated to v2.0 on ghcr.io:

https://github.com/kernelkit/infix/pkgs/container/infix-test/283471275?tag=2.0

ahmkar94:
> > I can checkout your branch and push a new infamy docker image.
> 
> Container image updated to v2.0 on ghcr.io:
> 
> https://github.com/kernelkit/infix/pkgs/container/infix-test/283471275?tag=2.0

Thnx!

troglobit:
> I meant that as you said _get is used by both get() and get_config(), but _get does not allow for the `source` parameter to be passed when get_config is used as a getter function. This way get_config will always fetche the default (running configuration) when calling it (as far as I understand) :)

OK.  Sometimes upgrading/replacing our own wrappers is not a 1:1 mapping to new upstream APIs, this may lead to more or less work to refactor in several steps.

troglobit:
> But I can make an additional commit for that anyways and we see then :)

Yes, please submit a proposal. Much easier to discuss around code :-)

test/.env
mattiaswal: I already pushed 2.0, i think we should change this for the future.

test/infamy/netconf.py
ahmkar94: Hmmm, I don't see any complaint in the test logs

test/infamy/netconf.py
ahmkar94: removed the extra whitespace

troglobit (approved):
Nicely done! 😎👍

PR #696, opened 2024-10-09
Minor fixes for the R2S and similar boards

troglobit:
## Description

Two minor fixes to:

 1. reduce log noise
 2. fix "show interfaces" on NICs that don't support reading RMON counters in JSON format

## Checklist

- [X] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Build related changes

### DIFF
--- a/board/common/rootfs/usr/libexec/infix/init.d/20-nameif
+++ b/board/common/rootfs/usr/libexec/infix/init.d/20-nameif
@@ -30,12 +30,16 @@ for netif in /sys/class/net/*; do
 done
 
 # Find and mark all switch ports
-ports=$(devlink -j port | jq -r '.port[]
-				 | select(.flavour == "physical")
-				 | .netdev')
-for iface in $ports; do
+if command -v devlink > /dev/null 2>&1; then
+    ports=$(devlink -j port | jq -r '.port[]
+				     | select(.flavour == "physical")
+				     | .netdev')
+    for iface in $ports; do
 	ip link set "$iface" group port
-done
+    done
+else
+    echo "Skipping switch port check, no devlink support."
+fi
 
 # At least loopback in iface group
 ip link set lo group iface
--- a/src/statd/python/yanger/yanger.py
+++ b/src/statd/python/yanger/yanger.py
@@ -726,12 +726,15 @@ def add_ip_addr(ifname, iface_in, iface_out):
 
 
 def add_ethtool_groups(ifname, iface_out):
-    """Fetch interface counters from kernel"""
-
-    data = run_json_cmd(['ethtool', '--json', '-S', ifname, '--all-groups'],
-                        f"ethtool-groups-{ifname}.json")
-    if len(data) != 1:
-        logger.warning(f"{ifname}: no counters available, skipping.")
+    """Fetch interface counters from kernel (need new JSON format!)"""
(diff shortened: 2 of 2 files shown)


board/common/rootfs/usr/libexec/infix/init.d/20-nameif
```diff
@@ -30,12 +30,16 @@ for netif in /sys/class/net/*; do
 done
 
 # Find and mark all switch ports
-ports=$(devlink -j port | jq -r '.port[]
-				 | select(.flavour == "physical")
-				 | .netdev')
-for iface in $ports; do
+if command -v devlink > /dev/null 2>&1; then
```
wkz, line 33: Do we know why `devlink` is not installed on these targets? Presumably the same version of `iproute2` is used as on `aarch64_deconfig`

wkz (commented):
Something irks me about having all these tests for various commands scattered across the code base.

I'm fine with having these semi-experimental `defconfig`s for various boards, but I'm not sure about complicating the main code to avoid a log message that is only ever emitted on those boards. I think a better approach would be to say: If you don't want those messages, build all the software needed to run Infix in full.

What do you think?

src/statd/python/yanger/yanger.py
```diff
@@ -726,12 +726,15 @@ def add_ip_addr(ifname, iface_in, iface_out):
 
 
 def add_ethtool_groups(ifname, iface_out):
-    """Fetch interface counters from kernel"""
-
-    data = run_json_cmd(['ethtool', '--json', '-S', ifname, '--all-groups'],
-                        f"ethtool-groups-{ifname}.json")
-    if len(data) != 1:
-        logger.warning(f"{ifname}: no counters available, skipping.")
+    """Fetch interface counters from kernel (need new JSON format!)"""
```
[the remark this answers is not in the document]
troglobit: Correct, I unfortunately had to explicitly disable it in d2457ba because otherwise `ethtool` does not work at all.

board/common/rootfs/usr/libexec/infix/init.d/20-nameif
troglobit: Good point, I realize now it actually is installed, so it must be the kernel side that is missing here too.

troglobit:
> [snip] I'm not sure about complicating the main code to avoid a log message that is only ever emitted on those boards. I think a better approach would be to say: If you don't want those messages, build all the software needed to run Infix in full.

Fair point, I'll do a retake on the changes and submit a new one at a later time.

board/common/rootfs/usr/libexec/infix/init.d/20-nameif
troglobit: Looked at the wrong target (the r2s).  This fix was for a log warning I ran into on `x86_64*_defconfig` -- it does not have `devlink` because it doesn't have LIBMNL enabled.  LIBMNL is selected on r2s and other aarch64 targets by the mdio-tools package, which we don't select on x86_64 ... sneaky.

So we should probably just add LIBMNL to the x86_64 targets.

troglobit:
New PR for these issues in #699.

PR #716, opened 2024-10-15
confd: fix double free of container hostname

rical:
## Description
Fix a double free in the memory minefield surrounding container hostname provisioning. The hostnamefmt() function already does free() on success.

This caused all sorts of mayhem when working with multiple containers. However it mainly manifested itself by some containers not starting due to there activation script being polluted with gibberish name data. Your symptom might be different..

### Example of polluted activation script
``container --hostname foobar --net veth0k -r always \
  create ^S@`M-}^G docker://ghcr.io/kernelkit/curios:edge``

## Side note
I haven't looked into it but is it really legit to free memory returned from the various lydx_ functions used here?
It seems to work as intended but it looks risky at a glance.

## Checklist

- [x] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes

### DIFF
--- a/src/confd/src/infix-containers.c
+++ b/src/confd/src/infix-containers.c
@@ -52,12 +52,10 @@ static int add(const char *name, struct lyd_node *cif)
 	if ((string = lydx_get_cattr(cif, "hostname"))) {
 		char *fmt = (char *)string;
 
-		if (hostnamefmt(&confd, &fmt)) {
+		if (hostnamefmt(&confd, &fmt))
 			ERRNO("%s: failed setting custom hostname", name);
-		} else {
+		else
 			fprintf(fp, " --hostname %s", fmt);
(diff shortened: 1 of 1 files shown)


troglobit:
Ouch, nice catch!  Although I cannot cannot reproduce in my limited setup, I agree that the current code needs work, and your fix is a good start, so let's merge this.

I'll submit a PR later that refactors `int hostnamefmt(struct confd *confd, char **fmt)` into `int hostnamefmt(struct confd *confd, const char *fmt, char *buf, size_t len)`.  Because we know that a hostname in Linux can only be 64 chars long, so a stack buffer for the return value is a much safer alternative.

PR #749, opened 2024-10-21
Disable BusyBox Related NTP Services

axkar:
When it comes to NTP services, Infix completely relies on Chrony. All other (BusyBox) NTP services should be disabled

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes


troglobit (changes requested):
Sooo sorry, I realize now I should have explained this task a little bit more in detail! :disappointed_relieved: 

What you need to do is run:

    make busybox-menuconfig

disable relevant parts for the keyword 'NTP'.  (Seach by tapping '/') And the save and exit busybox's menuconfig.  Then you call 

    make busybox-update-config

Which tells Buildroot to update the common BusyBox defconfig, somewhere down in `board/common/`.  Check with `git status` and verify with `git diff` that only `*NTP*` related changes were made.

----

### Background

BusyBox is a collection of small implementations of common UNIX/Linux utilities, commands, and daemons (services).  Usually built as a single "multicall binary"[^1]

So you actually don't have to touch these files in your current PR at all.  The whole `skeleton-init-finit` package actually adapts to other package selections in Buildroot and Infix as a whole.  (I periodically synchronize it with https://github.com/troglobit/finit-skel)

[^1]: A multicall binary has all tools built into one, and if you create a symlink to `/usr/bin/busybox`, e.g., `ls` and then call `ls`, the busybox binary detects it's called with another name and, in this case, calls ls_main().  This saves a LOT of space in embedded systems since you only need one ELF header for the single binary.  In fact, many embedded systems can make do with just that on top of an equally stripped Linux kernel.

troglobit (approved):
Awesome, thank you!

PR #764, opened 2024-10-24
Simplify Running Infix Tests from External Projects

### LINKED ISSUE #763: Simplify running infix tests from external projects
Currently, running Infix tests from an external project requires specifying the path to the topology.dot.in file and providing the image name.
This should be simplified.

axkar:
This will facilitate running infix tests from external projects:
  - eliminate the need for an external topology.dot.in file
  - eliminate the requirement to manually provide external image name

Fixes #763

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [x] Test changes
- [ ] Build related changes

### DIFF
--- a/test/test.mk
+++ b/test/test.mk
@@ -13,18 +13,16 @@ GIT_PATH            = $(BR2_EXTERNAL_INFIX_PATH)
 endif
 GIT_VERSION         = $(shell git -C $(GIT_PATH) describe --dirty --always --tags)
 
-IMAGE ?= infix
-TOPOLOGY-DIR ?= $(test-dir)/virt/quad
-
 base := -b $(base-dir)
 
 TEST_MODE ?= qeneth
-mode-qeneth := -q $(TOPOLOGY-DIR)
+mode-qeneth := -q $(test-dir)/virt/quad
 mode-host   := -t $(or $(TOPOLOGY),/etc/infamy.dot)
 mode-run    := -t $(BINARIES_DIR)/qemu.dot
 mode        := $(mode-$(TEST_MODE))
 
-binaries-$(ARCH) := $(addprefix $(IMAGE)-$(ARCH),.img -disk.img .pkg)
+INFIX_IMAGE_ID := $(call qstrip,$(INFIX_IMAGE_ID))
+binaries-$(ARCH) := $(addprefix $(INFIX_IMAGE_ID),.img -disk.img .pkg)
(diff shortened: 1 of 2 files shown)


test/test.mk
```diff
@@ -13,9 +13,7 @@ GIT_PATH            = $(BR2_EXTERNAL_INFIX_PATH)
 endif
 GIT_VERSION         = $(shell git -C $(GIT_PATH) describe --dirty --always --tags)
 
-IMAGE ?= infix
-TOPOLOGY-DIR ?= $(test-dir)/virt/quad
-
+TOPOLOGY-DIR := $(test-dir)/virt/quad
```
wkz, line 16: Since this does not need to be changed anymore, maybe we can remove it and just have...

```make
mode-qeneth := -q $(test-dir)/virt/quad
```

...below?

For future reference: I also like to put variables that can be overridden by the user in all-caps, but I will always `SNAKE_CASE` these, rather than `KEBAB-CASING` them. This makes it easier to pass them from a shell, since a dash is not a valid character in a  _shell_ variable, while it is in a _make_ variable.

wkz (approved):
So great that upper trees won't have to think about this at all anymore!

I added some minor comments. Your choice whether you fix them or not :smile:

test/test.mk
```diff
@@ -24,7 +22,8 @@ mode-host   := -t $(or $(TOPOLOGY),/etc/infamy.dot)
 mode-run    := -t $(BINARIES_DIR)/qemu.dot
 mode        := $(mode-$(TEST_MODE))
 
-binaries-$(ARCH) := $(addprefix $(IMAGE)-$(ARCH),.img -disk.img .pkg)
+INFIX_IMAGE_ID := $(subst ",,$(INFIX_IMAGE_ID))
```
wkz, line 25: Please use Buildroot's `qstrip` which takes care of any leading and/or trailing whitespace:

```suggestion
INFIX_IMAGE_ID := $(call qstrip,$(INFIX_IMAGE_ID))
```

test/test.mk
axkar: Yes, I'll fix that :)
Thnx

test/test.mk
axkar: Sure :)

Issue #751, opened 2024-10-22
R2S rousette crash on boot if LAN not connected


minexn:
### Current Behavior

The system booted with WAN connected and LAN disconnected.

```
lan             ethernet   DOWN        b2:b2:41:23:71:82                        
                ipv4                   192.168.2.1/24 (static)
wan             ethernet   UP          b2:b2:41:23:71:83                        
                ipv4                   10.10.3.101/24 (dhcp)
```

Rousette keeps crashing

```
Oct 22 06:47:02 r2s rousette[1883]: [2024-10-22 06:47:02.245] [rousette] [info] NACM config validation: Anonymous user access disabled 
Oct 22 06:47:02 r2s rousette[1883]: [2024-10-22 06:47:02.248] [rousette] [warning] Telemetry disabled. No CzechLight YANG modules found. 
Oct 22 06:47:07 r2s finit[1]: Service rousette keeps crashing, not restarting.
```

The system booted with WAN and LAN connected.
```
lan             ethernet   UP          b2:b2:41:23:71:82                        
                ipv4                   192.168.2.1/24 (static)
wan             ethernet   UP          b2:b2:41:23:71:83                        
                ipv4                   10.10.3.101/24 (dhcp)
```
Rousette starts and responds to queries.
```
Oct 22 06:49:09 r2s rousette[1542]: [2024-10-22 06:49:09.645] [rousette] [info] NACM config validation: Anonymous user access disabled 
Oct 22 06:49:09 r2s rousette[1542]: [2024-10-22 06:49:09.662] [rousette] [warning] Telemetry disabled. No CzechLight YANG modules found. 
```

### Expected Behavior

Rousette starts and responds to queries.
```
Oct 22 06:49:09 r2s rousette[1542]: [2024-10-22 06:49:09.645] [rousette] [info] NACM config validation: Anonymous user access disabled 
Oct 22 06:49:09 r2s rousette[1542]: [2024-10-22 06:49:09.662] [rousette] [warning] Telemetry disabled. No CzechLight YANG modules found. 
```

### Steps To Reproduce

load v24.10.1
unplug LAN
reboot
check log

### Additional information

Factory configuration

troglobit:
Reproduced on my R2S:

```
Oct 23 05:35:27 r2s finit[1]: Starting rousette[2163]
Oct 23 05:35:27 r2s rousette[2163]: [2024-10-23 05:35:27.538] [rousette] [info] NACM config validation: Anonymous user access disabled 
[...]

PR #768, opened 2024-10-26
nghttp2-asio: resolve numeric IPv6 server address

sgsx3:
The boost library refuses to resolve a numeric IPv6 host (::1) because its resolver flags are set to 'address_configured' by default. This patch simply runs an additional query in such a case with flags set to 'numeric_host'.

See https://www.boost.org/doc/libs/1_83_0/doc/html/boost_asio/reference/ip__resolver_base.html for more info.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes

### DIFF
--- a/patches/nghttp2-asio/0001-resolve-numeric-ipv6.patch
+++ b/patches/nghttp2-asio/0001-resolve-numeric-ipv6.patch
@@ -1,0 +1,18 @@
+diff --git a/lib/asio_server.cc b/lib/asio_server.cc
+index 74c9227..c3d42aa 100644
+--- a/lib/asio_server.cc
++++ b/lib/asio_server.cc
+@@ -82,8 +82,13 @@ boost::system::error_code server::bind_and_listen(boost::system::error_code &ec,
+   // Open the acceptor with the option to reuse the address (i.e.
+   // SO_REUSEADDR).
+   tcp::resolver resolver(io_service_pool_.get_io_service());
++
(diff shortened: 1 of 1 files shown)


mattiaswal (commented):
Thanks for your contribution! This bug really haunts me on the nights, if this is the fix, why does it work for us to resolve with our aarch64 platform (and on x86). 

So much unclear right now, have you reproduced it on x86 (Virtual)?

But during the exploring of this bug we at least found a related bug:
the /etc/hosts file does not mention ::1 (as it should)

Just some small things:
- Add fix #751 in your PR description.

troglobit:
Agreed, very weird that it works in our primary use-cases when we completely remove all IPv6 addresses except the one on loopback :thinking: 

@sgsx3 please also update the ChangeLog, and tick the boxes in the checklist.  Maybe a bit formal, but we've tried to make it so we can follow changes back to issues when looking for future bugs (answers the question: "*why* is this code here?").  We've tried to outline this in the [Contributing Guidelines](https://github.com/kernelkit/infix/blob/main/.github/CONTRIBUTING.md).  Thanks!

mattiaswal:
Does it work if you use an IPv4 address, (change in rousette main.cpp to 127.0.0.1 or localhost)

PR #799, opened 2024-11-04
test: Multi-Port Network Namespaces

### LINKED ISSUE #261: infamy: Add support for link breakers
To be able to test link loss without link down, required to test LACP, RSTP, MRP and BFD.

An example can be to add an attribute to a link in the physical topology:
<pre>
dut3:eth2 -- dut1:eth2 [controllable=qemu] # qemu monitor / a controllable switch when test physical devices
</pre>

And in the virtual topology:
dut2:ring1 -- dut3:ring2 [breakeable] <-- indicates that the link needs to be able to have a link breaker

wkz:
## Description

- Generalize the existing `IsolatedMacVlan` class to support creating namespaces with multiple ports
- Provide a Two-Port MAC Relay (TPMR) as an example implementation, which by default just relays frames between two ports, but can dynamically be set to block frames from passing through it. This is useful to test fail-over behavior of various protocols
- Add a test of BFD fail-over in combination with OSPF, using the new TPMR device

In order for the TPMR to act as a bridge, the mode of the MACVLANs has been changed to `passthru`, which allows it to receive all packets from its lower interface, and also lets it transmit unicast packets with any SA without triggering the kernel's spoof protection logic. This also enables us to use the parent interface's `promiscuity` value as an indicator for when a MACVLAN has been fully removed in the kernel, which should be more reliable than the old `time.sleep(.5)`.

Close #261 

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [x] Test changes
- [ ] Build related changes

### DIFF
--- a/test/infamy/netns.py
+++ b/test/infamy/netns.py
@@ -360,3 +360,39 @@ class Pcap:
                                  stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
                                  text=True, check=True)
         return tcpdump.stdout
+
+class TPMR(IsolatedMacVlans):
+    def __init__(self, a, b):
+        super().__init__(ifmap={ a: "a", b: "b" }, lo=False)
+
+    def start(self, forward=True):
+        ret = super().start()
+
+        for dev in ("a", "b"):
+            self.run(f"ip link set dev {dev} promisc on up".split())
+            self.run(f"tc qdisc add dev {dev} clsact".split())
+
+        if forward:
+            self.forward()
+
+        return ret
+
+    def _clear_ingress(self, iface):
+        return self.run(f"tc filter del dev {iface} ingress".split())
+
+    def _add_redir(self, frm, to):
+        cmd = \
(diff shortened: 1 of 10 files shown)


mattiaswal (approved):
Flawless as always :100:

test/infamy/netns.py
```diff
@@ -272,3 +360,39 @@ def tcpdump(self, args=""):
                                  stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
                                  text=True, check=True)
         return tcpdump.stdout
+
+class TPMR(IsolatedMacVlans):
+    def __init__(self, a, b):
```
troglobit, line 365: Would've been nice to have a class description here as well, just like you did for IsolatedMacVlan.

troglobit (approved):
Very nice, great to have a pure BFD test, finally. Only a minor comment.

PR #820, opened 2024-11-12
Ospf container

### LINKED ISSUE #652: Test a larger area 0.0.0.0 topology
4 duts, full mesh, count hops and verify neighbors, this to verify #635 

One more thing that our current testing does not cover is _special_ routes (black holes and more) this should also be tested. Maybe redistribute static and redistribute origin default (redistribute a default route)

### LINKED ISSUE #743: test: verify custom container interface
Some customers are using container interfaces with custom interface names (inside the container) and with a custom MAC address based on the chassis mac, with an offset 06:00:00:00:00:00.

 - Create two bridges (generic, non-VLAN filtering is fine), named `br0` and `br1`
 - Use two veth pairs `veth0a<-->veth0b`, and `veth1a<-->veth1b`
 - Set custom _container network_ interface names of `veth0a` and `veth1a`, use `br0` and `br1`, respectively (yes, really)
 - Set custom MAC address of `veth0a` (chassis + offset as above)
 - Set `veth0b` and `vet1b` as bridge ports in `br0` and `br1`, respectively
 - Add IP addresses on all interfaces (except the bridge ports, of course) to verify connectivity
 - Question: how to verify custom name when interface is inside container?
 - Maybe it would be useful in this test to also verify that disabling the container, and thus telling `podman` to move the interfaces back to the host's network namespace, also works?  Renaming the interface back to their original names without clashing with the bridges named with the same names as the interfaces inside the container.

Add an ASCII drawing (or svg) to this test so to illustrate the connections.  Preferably something prettier than the below:

                                      container
                                    _____________
                                   |             |
                                   |   br1  br0  |
               br0       br1       |___/___/_____|
                 \         \   vet1   /   /
                  \         \________/   /
                   \   veth0            /
                    \__________________/

> **Note:** it's quite possible this test cannot currently be automated since there is no way for *Infamy* to connect to the the container.  (The bundled container images do not have an SSH daemon and we don't want to go down the path of scripting the shell!)

### LINKED ISSUE #776: test: ospf-container-basic
### Purpose

This is a use-case test of multiple Infix features.  Figure 1 shows an overview of the test setup, each DUT is called R1 through R3.  The *Controller*, *ABR*, and `data` connections are simulated by the test PC.  The `ringN`  ports are connected to other DUTs via the test PC, which can act as a *link breaker*.

Image
_**Figure 1:** Topology overview._

Each DUT (R1-R3) contains, and verifies:

 - IP forwarding
 - OSPFv2 (IPv4)
 - Multiple containers'
   - two with dual networks
   - one container in *network host mode* to directly manipulate the host's firewall
 - Multiple network types

Image
_**Figure 2:** Single DUT with internal networking to containers, notice firewall container._

The DUTs are connected in a routed topology inside their own OSPF area.  A single area border router (ABR) is used to access the controller network in OSPF area 0.  Each router also provides "test points", connections where the controller (test PC) can attach other than its connection in area 0:

 - The *ringN* ports are intended to be connected to neighboring DUTs, but may at each end of the bus be used as test points
 - The *data* ports are intended to be test points for verifying connectivity with *container B* via *br1*
 - The *uplink* ports are for connecting to the *ABR*, at least one of the DUTs should *not* have a connection to the ABR, this to verify routing via another DUT

### Topology

Figure 1 shows the proposed test topology, excluding ABR and controller.  The Test PC emulates the ring, ensuring only tagged VLAN 8 traffic is forwarded between the DUTs.  Untagged traffic can be injected by the Test PC for traffic going to `br0` on each DUT.

### Stages

 - Connect from controller to each DUT's two containers A and B using the `uplink` connection
   - Verify HTTP access to container A
   - Verify SSH access to container B
 - Connect to `data` port of container B
   - Verify access to 192.0.2.1/24 using SSH from test PC
 - Verify forwarding of syslog traffic to controller
   - Inject untagged syslog UDP traffic to br0 on a DUT using one of the ring ports
   - Verify reception of NAT:ed syslog traffic

mattiaswal:
This fixes #652 (but larger 0.0.0.1 case is the same), fixes #776 and fixes #743
## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [x] Test changes
- [ ] Build related changes


troglobit (approved):
Nice work!  I like how the Frr patch(es) are now with the Frr package instead of in our Infix tree.  

As we agreed on AFK, we'll merge despite the issues we've seen on one test system.

wkz (approved):
Excellent! :1st_place_medal:

Issue #835, opened 2024-11-26
New container tests fail intermittently on Aarch64


troglobit:
The new use-case tests `ospf-container` and `container-firewall-basic` fail intermittently on Aarch64 test systems.  The failures result in "inconsistent container state" errors from Podman and manifest themselves as failure to  create/start at least one of the containers.

Further investigation show that the speed at which configuration is done at, using any of the M2M interfaces, unearths flaws in the asynchronous way that container creation/deletion is handled in the system.  When a user adds/deletes a container the actual job is enqueued to `execd`, which may lead to new configurations being applied before `execd` has completed a previous one.

This requires a redesign/simplification to ensure the M2M interfaces wait for, at least, the deletion and cleanup of containers before applying a new configuration.

PR #843, opened 2024-11-28
Redesign and simplify container creation/removal

troglobit:
## Description

The main purpose of this PR is to fix the unreliable behavior seen in #835, a few other bugs were also hunted down and fixed in the process: 

 - Pager does not exit if content fits on first screen, e.g., `show containers`
 - Avahi (mDNS) spams logs ... now silenced
 - Zeroconf agent not stopped when interfaces are removed
 - Finit template fix, e.g., creating another container marks all other containers "dirty"

## Checklist

- [X] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes


src/confd/src/infix-containers.c
```diff
@@ -191,55 +194,27 @@ static int add(const char *name, struct lyd_node *cif)
...
-	 * before being enqueued to execd, so we postpone it using this
-	 * "inbox" to the post hook.
-	 */
-	writesf(script, "a", "%s", _PATH_INBOX);
+	systemf("initctl -bnq touch container@%s.conf", name);
 
 	return 0;
 }
 
 static int del(const char *name)
```
wkz, line 207: This is not necessarily new behavior in this series, but I think the fact that we treat "no enabled" in the same way as removing the container as the same state might be problematic.

AFAIK, that means we will also remove any associated volumes, which I don't think is what the user (or yours truly) expects.

wkz (changes requested):
All hail @troglobit - destroyer of complexity! :laughing: 

Just want to check the "no enable" behavior before approving.

package/execd/tmpfiles.conf
```diff
@@ -1,7 +1 @@
-d  /run/containers/args		0700 - -
-d  /run/containers/files	0700 - -
-d  /var/lib/containers/oci	0755 - -
-d  /run/containers/inbox	0700 - -
 d  /run/containers/queue	0700 - -
```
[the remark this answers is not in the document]
troglobit: Had initially planned to drop it completely, but then decided to keep it in the tree for now.

src/confd/src/infix-containers.c
troglobit: Wow, good catch!  You are absolutely right, this is old buggy behavior.  Reworking to instead disable the container service, keeping any created container and volumes associated with it so they are not auto-pruned.

wkz (approved):
Perfect! :medal_sports:

PR #882, opened 2025-01-09
confd: Enable config for lldp tx interval

axkar:
This change introduces the configuration of the tx-interval parameter for the LLDP service, allowing control over the frequency of LLDP hello messages. By reducing the tx-interval, the waiting time for LLDP packets during tests is significantly shortened. This improvement eliminates the need to toggle the interface down and up to force the emission of LLDP packets.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes


src/confd/yang/infix-lldp.yang
```diff
@@ -9,7 +9,11 @@ module infix-lldp {
 
   organization "KernelKit";
   contact      "kernelkit@googlegroups.com";
-  description  "Infix augments and deviations to ieee-dot1ab-lldp.";
+  description  "This module extends and modifies the ieee802-dot1ab-lldp YANG model to add specific configuration options.";
```
troglobit, line 12: Please revert this.  The description is just longer and rambling.

src/confd/yang/infix-lldp.yang
```diff
@@ -21,11 +25,20 @@ module infix-lldp {
...
       description "Globally enable or disable IEEE 802.1ab LLDP agent.";
     }
+
+    leaf tx-interval {
+      type uint16 {
+        range "1..3600";
+      }
+      default "30";
+      description "Specifies the regular interval (in seconds) for LLDP message transmissions.";
+    }
```
troglobit, line 41: We should not add our own `tx-interval` when there is a `message-tx-interval` in the standard model.  See `grouping lldp-cfg` for details.

troglobit (changes requested):

src/confd/yang/infix-lldp.yang
axkar: Sure :) 
But sounds a bit weird "Infix augments and deviations".
Maybe we just need to turn "augments" to "augmentations" :) ?

axkar: Or something like "Infix augments and deviates from ieee-dot1ab-lldp"

troglobit: The original description is perfectly valid english, 'augments' is the plural of 'augment', see https://www.merriam-webster.com/dictionary/augment

src/confd/yang/infix-lldp.yang
troglobit: The guiding idea is that we should always use standard YANG models, unless they are very difficult or impossible to map to Linux or standard services (UNIX daemons) we use.  

For the LLDP yang model I've temporarily disabled all nodes and added a global `enabled` setting, which was missing.  The rest of the model should be used since `lldpd` is otherwise modeled exactly on the IEEE standard.

The `unit "ticks"` here is something you need to check against both the lldpd implementation and the IEEE standard.  It could mean 1 tick == 10 milliseconds, but since the default is 30 in the standard, and lldpd seems to use a default transmit interval of 30 seconds I'd venture to say that it may be seconds.  But please dive into that.  You can download the standard for free to verify.

src/confd/yang/infix-lldp.yang
axkar: Upps, my deep apology. I'll definitely revert the line :)

src/confd/yang/infix-lldp.yang
axkar: I'll check that :)

src/confd/src/infix-services.c
```diff
@@ -299,13 +299,41 @@ static int mdns_change(sr_session_ctx_t *session, uint32_t sub_id, const char *m
...
+		if (systemf("initctl -nbq %s lldpd", lldp_enabled ? "enable" : "disable"))
+			ERROR("Failed %s lldpd", lldp_enabled ? "enabling" : "disabling");
+		
+		if (lldp_enabled)
+			systemf("initctl -nbq touch lldpd");
+	}
+
+	if (lldp_enabled) {
+		const char *tx_interval = lydx_get_cattr(srv, "message-tx-interval");
+		if (tx_interval && *tx_interval && systemf("sudo lldpcli configure lldp tx-interval %s", tx_interval))
```
troglobit, line 332: Minor, should be one empty line between variable declarations and code. (Linux coding style)

The `message-tx-interval` is guaranteed to always have a value in the YANG model.  So you should not need to check the pointer at all.

Also, what happens if `lldpd` is not running when `lldpcli` is called above?  Hint: you probably need to write this configuration to the config file as well, and then "touch" the Finit service to ensure `lldpd` is reloaded when it is running.

troglobit (changes requested):
Nice work so far!

src/confd/src/infix-services.c
axkar: Thnx. I hope the new version looks better :)

src/confd/src/infix-services.c
```diff
@@ -299,13 +299,49 @@ static int mdns_change(sr_session_ctx_t *session, uint32_t sub_id, const char *m
...
+		if (!fp) {
+			ERROR("Failed to open /etc/lldpd.d/tx-interval.conf for writing");
+			return put(cfg);
+		}
+		fprintf(fp, "configure lldp tx-interval %s\n", tx_interval);
+		fclose(fp);
+
+		if (systemf("initctl -nbq restart lldpd")) {
+			ERROR("Failed to restart lldpd");
+		}
```
troglobit, line 341: Please do not use `restart` here, we want to "roll in" the next generation at the same time via finit's `initctl reload` mechanism.  The reason for this is simply that changes we add to the `/etc/lldpd.d/*.conf` may include references to interfaces that have not been created yet.  So, instead of restart, use `initctl -nbq touch ...` construct, used for other services in this file, to *mark it* for restarting by Finit.

Now, previously we thought `lldpd` reconfigured itself automatically on SIGHUP, but that does not seem to be the case.  To ensure Finit does not send SIGHUP, but instead performs the actual process restart (stop + start) for us, we need to change the .conf snippet in `package/skeleton-init-finit/skeleton/etc/finit.d/available/lldpd.conf`.

From

    service [2345] env:-/etc/default/lldpd lldpd -d $LLDPD_ARGS -- LLDP daemon (IEEE 802.1ab)

to:

    service <!> env:-/etc/default/lldpd \
            [2345] lldpd -d $LLDPD_ARGS -- LLDP daemon (IEEE 802.1ab)

The `<!>` construct tells Finit this service does not support SIGHUP.

To test this, without rebuilding from distclean, you need to perform a `make skeleton-init-finit-rebuild`.

troglobit (changes requested):
Only one final comment, nice work!

src/confd/src/infix-services.c
axkar: Clear as a tear!

axkar:
@troglobit Thnx for all the comments so far. I made a major rework in the latest commit. Please have another look when you have time :)

src/confd/src/infix-services.c
```diff
@@ -299,13 +302,64 @@ static int mdns_change(sr_session_ctx_t *session, uint32_t sub_id, const char *m
...
+		if (!enabled) {
+			if (rmrf(LLDP_CONFIG))
+				ERROR("Failed to remove old %s: %d", LLDP_CONFIG, errno);
+		}
+		return SR_ERR_OK;
+	
+	case SR_EV_ABORT:
+		if (fexist(LLDP_CONFIG_NEXT))
+			rmrf(LLDP_CONFIG_NEXT);			
+
```
mattiaswal, line 355: Not releasing cfg.

src/confd/src/infix-services.c
```diff
@@ -299,13 +302,64 @@ static int mdns_change(sr_session_ctx_t *session, uint32_t sub_id, const char *m
...
 static int lldp_change(sr_session_ctx_t *session, uint32_t sub_id, const char *module,
 	const char *xpath, sr_event_t event, unsigned request_id, void *_confd)
-{
-	return svc_change(session, event, xpath, "lldp", "lldpd");
+{	
+	struct lyd_node * lldp_lyd_node = NULL;
+	sr_data_t *cfg;
+	int enabled;
+
+	if (sr_get_data(session, xpath, 0, 0, 0, &cfg) || !cfg)
```
mattiaswal, line 315: Move this to inside EV_CHANGE, it is the only place you use it.
remember to release it.

src/confd/src/infix-services.c
```diff
@@ -299,13 +302,64 @@ static int mdns_change(sr_session_ctx_t *session, uint32_t sub_id, const char *m
...
+		if (fexist(LLDP_CONFIG_NEXT))
+			if (fexist(LLDP_CONFIG))
+				if (rmrf(LLDP_CONFIG))
+					ERROR("Failed to remove old %s: %d", LLDP_CONFIG, errno);			
+
+		rename(LLDP_CONFIG_NEXT, LLDP_CONFIG);
+		
+		svc_change(session, event, xpath, "lldp", "lldpd");
+
+		if (!enabled) {
```
mattiaswal, line 346: Check if LLDPD_CONFIG_NEXT exist, then it is enabled.

src/confd/src/infix-services.c
axkar: I need enable flag in SR_EV_DONE to remove the config

src/confd/src/infix-services.c
```diff
@@ -299,13 +302,64 @@ static int mdns_change(sr_session_ctx_t *session, uint32_t sub_id, const char *m
...
+
+		FILE *fp = fopen(LLDP_CONFIG_NEXT, "w");
+		if (!fp) {
+			ERROR("Failed to open %s for writing", LLDP_CONFIG_NEXT);
+			return put(cfg);
+		}
+		fprintf(fp, "configure lldp tx-interval %s\n", tx_interval);
+		fclose(fp);
+		
+		return SR_ERR_OK;
```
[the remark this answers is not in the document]
axkar: Fixed

src/confd/src/infix-services.c
axkar: Fixed. Should be cleaner now

mattiaswal (approved):
Great work! :1st_place_medal:

src/confd/yang/infix-lldp.yang
```diff
@@ -11,6 +11,10 @@ module infix-lldp {
   contact      "kernelkit@googlegroups.com";
   description  "Infix augments and deviations to ieee-dot1ab-lldp.";
 
+  revision 2025-01-08 {
+    description "Switched to the base lldp:message-tx-interval";
```
troglobit, line 15: This should probably read, `"Enable std. /lldp:lldp/message-tx-interval"`, because we are not *switching from* anything but rather unlocking something that was locked-down/disabled before.

src/libsrx/src/helpers.c
```diff
@@ -35,7 +35,7 @@ static int do_delete(const char *fpath, const struct stat *sb, int tflag, struct
 
 int rmrf(const char *path)
 {
-	if (!fisdir(path))
+	if (!fexist(path))
```
troglobit, line 38: Why this change?  There is no mention of it in the commit message (should possibly be a separate commit) and the original intent of the `rmrf()` function was to remove an entire *directory* structure recursively.

To delete a single file use `remove()` or `unlink()` from the standard C library.  There are some convenience wrappers in libite called `erase()`, `erasef()`, and `fremove()` that can also be useful.  All of which are more suited to removing a single file which is all that `lldp_change()` needs.

src/confd/src/infix-services.c
```diff
@@ -299,13 +302,62 @@ static int mdns_change(sr_session_ctx_t *session, uint32_t sub_id, const char *m
...
+			fprintf(fp, "configure lldp tx-interval %s\n", tx_interval);
+			fclose(fp);
+		}
+			
+		return put(cfg);		
+
+	case SR_EV_DONE:
+		if (fexist(LLDP_CONFIG_NEXT)){
+			if (fexist(LLDP_CONFIG)){
+				if (rmrf(LLDP_CONFIG))
```
troglobit, line 337: Please use the libite helper function `erase()` here instead.  With that you can skip the call to `fexist()`, since `erase()` only fails if there was an error *other* than the file doesn't exist.

Apply this to the other constructs using `rmrf()` as well in this file.

troglobit (changes requested):
All in all very good job on this!  Some minor things left to change and revert, but then we can merge this.

An observation though, it's beginning to look like the code for the LLDP model needs to move to a dedicated `.c` file very soon :smiley:

src/confd/yang/infix-lldp.yang
axkar: I agree :)

src/libsrx/src/helpers.c
axkar: I'll go for libite erase() definitely :) Thnx

src/confd/src/infix-services.c
axkar: I'll do that :)

axkar:
@mattiaswal I've added a commit for port-enable feature. In case you would like to have a look :)

src/confd/src/infix-services.c
```diff
@@ -299,13 +302,71 @@ static int mdns_change(sr_session_ctx_t *session, uint32_t sub_id, const char *m
...
+			}
+			fclose(fp);
+		}
+
+		return put(cfg);
+
+	case SR_EV_DONE:
+		if (fexist(LLDP_CONFIG_NEXT)){
+			if (erase(LLDP_CONFIG))
+				ERROR("Failed to remove old %s", LLDP_CONFIG);
```
troglobit, line 350: For APIs that rely on POSIX we usually get `errno` set when the API fails.   That is why we have the `ERRNO()` macro that takes the same argument as `ERROR()` but appends a human-friendly error message as well.

src/confd/src/infix-services.c
```diff
@@ -299,13 +302,71 @@ static int mdns_change(sr_session_ctx_t *session, uint32_t sub_id, const char *m
...
+
+			rename(LLDP_CONFIG_NEXT, LLDP_CONFIG);
+		}
+		else
+			if (erase(LLDP_CONFIG))
+				ERROR("Failed to remove old %s", LLDP_CONFIG);
+
```
troglobit, line 357: Same here, please use `ERRNO()` macro.

src/confd/src/infix-services.c
```diff
@@ -299,13 +302,71 @@ static int mdns_change(sr_session_ctx_t *session, uint32_t sub_id, const char *m
...
+		}
+		else
+			if (erase(LLDP_CONFIG))
+				ERROR("Failed to remove old %s", LLDP_CONFIG);
+		
+		svc_change(session, event, xpath, "lldp", "lldpd");
+		break;
+
+	case SR_EV_ABORT:
+		erase(LLDP_CONFIG_NEXT);
```
troglobit, line 362: This is fine to ignore the result of, like you've done, since it's in our preparation path.

src/confd/src/infix-services.c
```diff
@@ -463,16 +524,13 @@ static int change_keystore_cb(sr_session_ctx_t *session, uint32_t sub_id, const
...
+		rmrf(SSH_HOSTKEYS_NEXT);
 		return SR_ERR_OK;
 	case SR_EV_DONE:
 		if(fexist(SSH_HOSTKEYS_NEXT)) {
-			if(fexist(SSH_HOSTKEYS))
-				if(rmrf(SSH_HOSTKEYS)) {
-					ERROR("Failed to remove old SSH hostkeys: %d", errno);
-				}
+			if(rmrf(SSH_HOSTKEYS)) {
+				ERROR("Failed to remove old SSH hostkeys: %d", errno);
```
troglobit, line 532: Forgot one `rmrf` -> `erase`, please also use `ERRNO()` here as well.

troglobit (changes requested):
Almooooost there! :smile:

src/confd/src/infix-services.c
troglobit: My bad, you're right this is a directory!  Just fix the ERROR -> ERRNO.

troglobit (changes requested):

src/confd/src/infix-services.c
axkar: Fixed

src/confd/src/infix-services.c
```diff
@@ -299,13 +302,71 @@ static int mdns_change(sr_session_ctx_t *session, uint32_t sub_id, const char *m
...
+	case SR_EV_CHANGE:
+		if (sr_get_data(session, xpath, 0, 0, 0, &cfg) || !cfg)
+			break;
+
+		node = cfg->tree;
+		if (lydx_is_enabled(node, "enabled")){
+			const char *tx_interval = lydx_get_cattr(node, "message-tx-interval");
+			FILE *fp = fopen(LLDP_CONFIG_NEXT, "w");
+			if (!fp) {
+				ERROR("Failed to open %s for writing", LLDP_CONFIG_NEXT);
```
troglobit, line 326: Sorry for a late review comment, but this should also be a `ERRNO()` instead of `ERROR()`.

src/confd/src/infix-services.c
```diff
@@ -299,13 +302,71 @@ static int mdns_change(sr_session_ctx_t *session, uint32_t sub_id, const char *m
...
+
+			LY_LIST_FOR(lyd_child(node), subnode) {
+    				if (!strcmp(subnode->schema->name, "port")) {
+					const char *port_name = lydx_get_cattr(subnode, "name");
+					const char *admin_status = lydx_get_cattr(subnode, "admin-status");
+
+					if (strcmp(admin_status, "tx-and-rx") == 0)
+						admin_status = "rx-and-tx";
+
+					fprintf(fp, "configure ports %s lldp status %s\n", port_name, admin_status);
```
troglobit, line 339: Nice catch with the `tx-and-rx` vs `rx-and-tx`!  I misread your code first :smiley::+1:

troglobit (changes requested):
Very nice, looks perfect!  I only had that minor thing with a lingering `ERROR()` that I missed in the first review round.

src/confd/src/infix-services.c
axkar: Thnx!

src/confd/src/infix-services.c
axkar: Ouch, thnx again. My bad ...

troglobit (approved):
Wohoo! :muscle::sparkles:

PR #883, opened 2025-01-09
Yanger Modularization

wkz:
## Description

Split up the `yanger.py` monolith into multiple Python submodules.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes


troglobit (approved):
This looks really good, it was high time to take care of the "`yanger` situation".

With the addition of more of the standard models (for linting) we can later also expand upon the ideas for `confd` to exclusively handle the installation and setup of sysrepo modules that is currently spread out over various packages.

Only one final note, I'm a bit apprehensive using the `ix` prefix too much.  For `ixll` it really fits since it's Infix device instances it operates on.  But the `ixyang` name feels a bit forced, it lives in the Infix `utils/` directory, so it's obviously an Infix command, so the `ix` prefix feels a bit redundant.  How about `op` or `run`, or just plain `yang`?

wkz:
> This looks really good, it was high time to take care of the "`yanger` situation".

Thanks! :heart:

> With the addition of more of the standard models (for linting) we can later also expand upon the ideas for `confd` to exclusively handle the installation and setup of sysrepo modules that is currently spread out over various packages.

:100: That would be awesome!

> Only one final note, I'm a bit apprehensive using the `ix` prefix too much. For `ixll` it really fits since it's Infix device instances it operates on. But the `ixyang` name feels a bit forced, it lives in the Infix `utils/` directory, so it's obviously an Infix command, so the `ix` prefix feels a bit redundant. How about `op` or `run`, or just plain `yang`?

Interesting, I made the exact opposite observation earlier today :rofl: 

I was thinking that we should maybe have a _more_ generic name for `ixll` (maybe `lldo` or something), since it is _not_ in any way limited to Infix peers.

`ixyang` on the other hand tries to help out with various tasks related to how _Infix_ manages _YANG_ models, so the name felt apt to me :man_shrugging: 

Could you expand a bit on what it is about the `ix`-prefix that rubs you the wrong way? I do not have a super strong opinion about this, but `op` and `run` is _very_ nondescript, no?

troglobit:
> > I'm a bit apprehensive using the `ix` prefix too much. For `ixll` it really fits since it's Infix device instances it operates on. But the `ixyang` name feels a bit forced, it lives in the Infix `utils/` directory, so it's obviously an Infix command, so the `ix` prefix feels a bit redundant. How about `op` or `run`, or just plain `yang`?

> Interesting, I made the exact opposite observation earlier today 🤣

Heh! :laughing: 

> I was thinking that we should maybe have a _more_ generic name for `ixll` (maybe `lldo` or something), since it is _not_ in any way limited to Infix peers.

That's a good point — `ixll` really is more generic than I first thought, and `lldo` feels like a better fit!

> `ixyang` on the other hand tries to help out with various tasks related to how _Infix_ manages _YANG_ models, so the name felt apt to me 🤷‍♂️
> 
> Could you expand a bit on what it is about the `ix`-prefix that rubs you the wrong way? I do not have a super strong opinion about this, but `op` and `run` is _very_ nondescript, no?

Well, a shorter name is quicker to type, and I’d like to avoid over-branding our tools — it can feel a bit forced or artificial.  The linting feature isn’t strictly Infix-specific, which makes the ix- prefix feel less natural to me.  I also see potential for this tool to grow, like adding a `grep` feature to find and show matching nodes.  I think the shorter `yang` is sufficient and leaves room for that kind of broader use.

PR #893, opened 2025-01-18
Add script for merging defconfigs

mattiaswal:
Useful to only create a new image with only small diff with for example aarch64_defconfig.

example:
If the only diff is the VENDOR, instead of KernelKit you want 'foobar'. create a file with the change:
INFIX_VENDOR="foobar"

and call ./utils/generate-defconfig.sh -b aarch64_defconfig -c foobar-diff.conf -o configs/foobar_defconfig

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes

### DIFF
--- a/utils/generate-defconfig.sh
+++ b/utils/generate-defconfig.sh
@@ -1,0 +1,80 @@
+#!/bin/sh
+# Create new defconfigs by only applying the changes from a standard defconfig
+set -e
+
+SCRIPT_DIR="$(readlink -f $(dirname -- "$0"))"
+MERGE_CONFIG="${SCRIPT_DIR}/../buildroot/support/kconfig/merge_config.sh"
+
+usage() {
+    cat <<EOF
+Usage: $(basename "$0") -b BASE_CONFIG -c CHANGES -o OUTPUT_FILE
+
+Create a new defconfig by applying changes to a base configuration.
+
+Required Options:
+    -b, --base BASE          Path to base defconfig file
+    -c, --changes CHANGES    Path to config changes to apply
+    -o, --output OUTPUT      Path to output defconfig file
+
+Example:
+    $(basename "$0") \\
+        -b configs/aarch64_defconfig \\
+        -c changes/no-containers.conf \\
+        -o configs/aarch64_wo_containers_defconfig
+
+EOF
+    exit 1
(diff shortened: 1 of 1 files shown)


utils/generate-defconfig.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/sh
+# Create new defconfigs by only applying the changes from a standard defconfig
+set -e
+
+SCRIPT_DIR="$( dirname -- "$0" )"
```
wkz, line 5: NIT:
```suggestion
SCRIPT_DIR="$(readlink -f $(dirname -- "$0"))"
```
Then you can remove the implicitly relative lookup when calling `./$MERGE_CONFIG` below

utils/generate-defconfig.sh
```diff
@@ -0,0 +1,80 @@
...
+    echo "Error: merge_config.sh not found or not executable at: $MERGE_CONFIG"
+    exit 1
+fi
+
+check_file "$base" "Base config"
+check_file "$changes" "Changes config"
+
+TMPDIR=`mktemp -d`
+./$MERGE_CONFIG -O "$TMPDIR" -n "$base" "$changes"
+O="$TMPDIR" make savedefconfig
```
wkz, line 78: I take it this only works when running from the root of a repo? Should we have a check for that?

utils/generate-defconfig.sh
```diff
@@ -0,0 +1,80 @@
...
+mv "$TMPDIR"/defconfig "output"
```
wkz, line 79: I would have been surprised if `generate-defconfig.sh` overwrote something in my build tree. But if that is the behavior we want, should we support the case (which many of us use) where this script is executed from _within_ an output tree?

wkz (changes requested):
Mmmm ~doughnuts~automation! :doughnut:

utils/generate-defconfig.sh
mattiaswal: There is a bug in the script, should be "$output", which is sent as input to the script. Not the output directory.

PR #901, opened 2025-01-22
Fix memory leaks

mattiaswal:
Also minor fixes, refactor syslog tests to use put_config_dicts and fix a minor bug in generate_defconfig script.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes

### DIFF
--- a/src/confd/src/ietf-keystore.c
+++ b/src/confd/src/ietf-keystore.c
@@ -118,8 +118,10 @@ static int change_cb(sr_session_ctx_t *session, uint32_t sub_id, const char *mod
 					goto next;
 				}
 			next:
-				unlink(SSH_PRIVATE_KEY);
-				unlink(SSH_PUBLIC_KEY);
+				if (erase(SSH_PRIVATE_KEY))
+					ERRNO("Failed removing SSH server private key");
+				if (erase(SSH_PUBLIC_KEY))
+					ERRNO("Failed removing SSH server public key");
 
 				if (priv_key)
 					free(priv_key);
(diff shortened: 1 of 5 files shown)


src/confd/src/ietf-keystore.c
```diff
@@ -118,8 +118,10 @@ static int change_cb(sr_session_ctx_t *session, uint32_t sub_id, const char *mod
 					goto next;
 				}
 			next:
-				unlink(SSH_PRIVATE_KEY);
-				unlink(SSH_PUBLIC_KEY);
+				if(fexist(SSH_PRIVATE_KEY))
+					unlink(SSH_PRIVATE_KEY);
+				if (fexist(SSH_PUBLIC_KEY))
+					unlink(SSH_PUBLIC_KEY);
```
troglobit, line 124: Since we don't check the return value of `unlink()` the new calls to `fexist()` is just waste.  

A better change may be to replace them with:

```c
if (erase(SSH_PRIVATE_KEY))
    ERRNO("Failed removing SSH server private key");
if (erase(SSH_PUBLIC_KEY))
    ERRNO("Failed removing SSH server private key");
```

The `ERRNO()` macro appends a human-friendly error message (mostly directed to us for troubleshooting).  We can rely on `erase()` to skip `ENOENT` errors (file does not exist), which we don't care about in this operation.

troglobit (changes requested):
Nice work!  Only found something minor to consider.

troglobit (approved):
:+1:

PR #907, opened 2025-01-23
Match on Edge Attributes when Mapping Topologies

### LINKED ISSUE #654: Add support for declaring and matching on edge attributes when mapping topologies
## Problem Description

In physical test rigs, it is typically not reasonable for the test host to provide the required number of Ethernet connections via dedicated NICs. Instead, an intermediate switch is deployed that will multiplex many ports to a single NIC on the host, by isolating each switch port in its own VLAN.

For most tests this setup is not an issue, but there are cases when the intermediate switch might not manage to act as a transparent multiplexer. Some typical examples:
1. IEEE Link-Local Multicast: Frames with a destination address in the` 01:80:c2:00:00:0x` range are not supposed to be forwarded, since they are link-local.
2. 802.1Q Tagged Frames: Some switches does not support assigning incoming traffic to the port's default PVID if the packet is already tagged.

## Proposed Solution

Add a comma-separated list of attributes that an edge provides, to physical topologies:

```dot
graph "my-rig" {
        node [shape=record];

        host [
	    label="host | { <d1a> d1a | <d1b> d1b | <d1c> d1c | <d1c> d1d | <d2a> d2a | <d2b> d2b | <d2c> d2c | <d2d> d2d | <d2a> d3a | <d3b> d3b | <d3c> d3c | <d3c> d3d |  <d4a> d4a | <d4b> d4b | <d4c> d4c | <d4c> d4d }",
	    kind="controller",
	];

        dut1 [
	    label="{ <e1> e1 | <e2> e2 | <e3> e3 } | dut1 | { <e4> e4 | <e5> e5 | <e6> e6 | <e7> e7 | <e8> e8}",
	    kind="infix",
	    ];
        dut2 [
	    label="{ <e1> e1 | <e2> e2 | <e3> e3 } | dut2 | { <e4> e4 | <e5> e5 | <e6> e6  | <e7> e7 | <e8> e8}",
	    kind="infix",
	];

	host:d1a -- dut1:e1 [kind=mgmt]
	host:d1b -- dut1:e2 [provides="ieee-mc-transparency,802.1q-transparency"]
	host:d1c -- dut1:e3
	host:d1d -- dut1:e4

	host:d2a -- dut2:e1 [kind=mgmt]
	host:d2b -- dut2:e2 [provides="ieee-mc-transparency,802.1q-transparency"]
	host:d2c -- dut2:e3
	host:d2d -- dut2:e4

	dut1:e5 -- dut2:e5
	dut1:e6 -- dut2:e6
}
```

In this example the controller has 4 connections to each DUT, where one is dedicated to management (`d<N>a`), one is a direct connection via a real NIC (`d<N>b`), and the other two are via an intermediate switch.

A test can then require that an edge provides these attributes via its logical topology. So a test that expects to sniff LLDP frames might say:

```dot
graph "lldp-basic" {
        node [shape=record];

	host [
	    label="host | { <tgt> tgt | <data> data }",
	    kind="controller",
	];

        target [
	    label="{ <mgmt> mgmt | <data> data } | target",
	    kind="infix",
	];

	host:tgt -- target:mgmt [kind=mgmt]
	host:data -- target:data [requires="ieee-mc-transparency"]
}
```

axkar:
Implemented logic to check if the system supports required services. Logical edges with 'requires' fields are mapped only if physical edges provide corresponding services. If not, the test is skipped. This ensures that tests only run when the required services are available in the test system configuration.

Fixes #654

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [x] Test changes
- [ ] Build related changes

### DIFF
--- a/test/infamy/topology.py
+++ b/test/infamy/topology.py
@@ -1,6 +1,8 @@
 import networkx as nx
 from networkx.algorithms import isomorphism
 
+from itertools import permutations
+import json
 
 def _qstrip(text):
     if text is None:
@@ -11,28 +13,26 @@ def _qstrip(text):
 
     return text
 
+def compatible(physical, logical):
+    provides_set = set(physical.get("provides", "").split())
+    requires_set = set(logical.get("requires", "").split())
+
+    return requires_set.issubset(provides_set)
 
 def map_edges(les, pes):
-    acc = []
-    les = sorted(list(les.values()), key=lambda x: x.get("kind", ""), reverse=True)
-    pes = sorted(list(pes.values()), key=lambda x: x.get("kind", ""), reverse=True)
+    les = les.values()
+    pes = pes.values()
 
-    for i in range(len(les)):
-        if pes[i].get("kind") != les[i].get("kind"):
-            return None
-
-        acc.append((les[i], pes[i]))
-
-    return acc
-
+    for perm in permutations(pes, len(les)):
+        candidate = tuple(zip(les, perm))
+        if all(map(lambda pair: compatible(pair[1], pair[0]), candidate)):
+            return candidate
 
 def match_node(pn, ln):
-    return pn.get("kind") == ln.get("kind")
-
+    return compatible(pn, ln)
 
 def match_edge(pes, les):
     return map_edges(les, pes) is not None
-
 
 class Topology:
     def __init__(self, dotg):
@@ -139,15 +139,16 @@ class Topology:
         return None
 
     def get_mgmt_link(self, src, dst):
-        return self.get_link(src, dst, lambda e: e.get("kind") == "mgmt")
+        return self.get_link(src, dst, lambda e: compatible(e, {"requires": "mgmt"}))
 
     def get_ctrl(self):
-        ns = self.get_nodes(lambda _, attrs: attrs.get("kind") == "controller")
+        ns = self.get_nodes(lambda _, attrs: compatible(attrs, {"requires": "controller"}))
         assert len(ns) == 1
         return ns[0]
 
     def get_infixen(self):
-        return self.get_nodes(lambda _, attrs: attrs.get("kind") == "infix")
+        return self.get_nodes(lambda _, attrs: compatible(attrs, {"requires": "infix"}))
+
 
     def get_attr(self, name, default=None):
         return _qstrip(self.dotg.get_attributes().get(name, default))
--- a/test/case/ietf_interfaces/bridge_basic/topology.dot
+++ b/test/case/ietf_interfaces/bridge_basic/topology.dot
@@ -9,16 +9,16 @@ graph "1x2" {
 	host [
 	    label="host | { <mgmt> mgmt | <data> data }",
 	    pos="0,12!",
-	    kind="controller",
+	    requires="controller",
 	];
 
         target [
 	    label="{ <mgmt> mgmt | <data> data } | target\n10.0.0.2/24 (br0)",
 	    pos="10,12!",
 
-	    kind="infix",
+	    requires="infix",
 	];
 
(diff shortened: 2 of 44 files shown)


test/infamy/topology.py
```diff
@@ -11,24 +13,36 @@ def _qstrip(text):
 
     return text
 
+def satisfies(provides, requires):
+    provides_set = set(provides.split() if isinstance(provides, str) and provides else provides)
+    requires_set = set(requires.split() if isinstance(requires, str) and requires else requires)
```
wkz, line 18:
```suggestion
    provides_set = set(provides.split())
    requires_set = set(requires.split())
```
Why is this not enough here? Can `provides` or `requires` ever _not_ be strings?

test/infamy/topology.py
```diff
@@ -11,24 +13,36 @@ def _qstrip(text):
...
+
+    return requires_set.issubset(provides_set)
```
wkz, line 20: So elegant when code almost turns into prose :nerd_face:

test/infamy/topology.py
```diff
@@ -11,24 +13,36 @@ def _qstrip(text):
...
+            if pe_tuple in used_pes or not satisfies(pe.get("provides", ""), le.get("requires", "")):
+                break
 
+            used_pes.add(pe_tuple)
+            acc.append((le, pe))
 
-def match_node(pn, ln):
-    return pn.get("kind") == ln.get("kind")
+        else:
+            return acc
```
wkz, line 40: I think this could be even simpler:

```python
for perm in permutations(pes, len(les)):
    candidate = tuple(zip(les, perm))
    if all(map(lambda le, pe: satisfies(pe.get("provides", ""), le.get("requires", "")), candidate)):
        return candidate
```

Also, if you were to push the lookup of `provides`/`requires` into the implementation of `satisfies()`, and swap the arguments around, i.e.:

```python
def satisfies(logical, physical):
    provides_set = set(physical.get("provides", "").split())
    requires_set = set(logical.get("requires", "").split())
    return requires_set.issubset(provides_set)
```

then it could be:

```python
for perm in permutations(pes, len(les)):
    candidate = tuple(zip(les, perm))
    if all(map(satisfies, candidate)):
        return candidate
```
...this would also get rid of a bunch of other `.get()`'s sprinkled throughout this change.

wkz (changes requested):
A-W-E-S-O-M-E job Ahmed! :muscle: This will allow us to keep dodging the dreaded manually-maintained-test-case-to-list-of-test-systems-filter-file :laughing: 

I think we can turn up the readability to :100: with some further improvements. Let me know what you think :+1:

test/infamy/topology.py
wkz: Looking at this again, I think `compatible()` might be a better name than `satisfies()` (which I am aware that I suggested in the first place :sweat_smile:)

test/infamy/topology.py
axkar: Thanx!

test/infamy/topology.py
axkar: It is actually enough. Thnx.
The whole complexity came out of the "requires" part sent from get_ctrl function. I was unnecessarily sending a list [ ] type instead of string, so had to support that afterwards.

test/infamy/topology.py
axkar: That was great!

axkar: Agree, renamed to compatible :)

wkz (approved):
Exquisite! :monocle_face: :tophat:

PR #946, opened 2025-02-16
confd: ensure admin users have full access to vtysh

troglobit:
## Description

Future Infix defconfigs for smaller system may exclude Frr and set all routes directly in the kernel.  So failing to add a user to the `frrvty` group should not cause an error, hence the new `group_exists()` function.

Note, the new `*_groups()` functions could also be used for `guest` and `operator` level users.  We've talked opening up guest access to sysrepo, relying fully on NACM, in which case we could create a new group `sysrepo` for `/dev/shm/*` files.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes
- [X] Other (please describe): convenience

### DIFF
--- a/src/confd/src/ietf-system.c
+++ b/src/confd/src/ietf-system.c
@@ -45,6 +45,13 @@ static char   *os  = NULL;
 static char   *nm  = NULL;
 static char   *id  = NULL;
 
+/* TODO: add `#ifdef HAVE_FOO` around optional features. */
+static const char *admin_groups[] = {
+	"wheel",
+	"frrvty",
+	NULL
+};
+
 static struct { char *name, *shell; } shells[] = {
 	{ "infix-system:sh",    "/bin/sh"    },
 	{ "infix-system:bash",  "/bin/bash"  },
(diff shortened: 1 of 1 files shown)


src/confd/src/ietf-system.c
```diff
@@ -45,6 +45,12 @@ static char   *os  = NULL;
 static char   *nm  = NULL;
 static char   *id  = NULL;
 
+static const char *admin_groups[] = {
+	"wheel",
+	"frrvty",
```
wkz, line 50:
```suggestion
#ifdef HAVE_FRR
	"frrvty",
#endif
```
As this list grows, having this type of structure might make it easier to keep track of which deps pull in which groups. It would also avoid the need for `group_exists()`. Just a thought.

Now that I think about it: Is there a `podman` group that we should add for builds with container support?

wkz (approved):
Thanks for fixing this! :1st_place_medal:

mattiaswal (approved):
Great 👌

src/confd/src/ietf-system.c
troglobit: Good point. I'll add that to my commit 😀👍

Not that I've seen, but I'll have a look again at podman. I'm still at it with the other related issues anyway.

PR #955, opened 2025-02-23
Lingering containers

### LINKED ISSUE #902: Lingering container instances are sometimes left on the system
### Current Behavior

In during [regression testing](https://gist.github.com/ael-bot/dc682b1a78f2613f657bdb5e34d1f9e9), a few tests failed because `yanger` did not handle the containers that were created _but not running_. That issue will be fixed separately, but it also prompted this issue: Namely:

Sometimes, even though no containers are configured on the system, created instances (presumably from some earlier test run) still remain in the system.

Here we see that `confd` reports no containers being configured, while `podman` lists three created instances:
```
root@C:~$ sysrepocfg -f json -X -d running -m infix-containers
{

}
root@C:~$ podman ps --all
CONTAINER ID  IMAGE                                                COMMAND               CREATED       STATUS      PORTS                  NAMES
584a280b9070  localhost/curios-nftables-oci-arm64-v24.11.0:latest  /usr/sbin/nft-hel...  17 hours ago  Created                            firewall
266b16ae8b0d  localhost/curios-httpd-oci-arm64-v24.11.0:latest     /usr/sbin/httpd -...  17 hours ago  Created     0.0.0.0:36959->91/tcp  container-A
cc33eb4b50a2  localhost/curios-httpd-oci-arm64-v24.11.0:latest     /usr/sbin/httpd -...  17 hours ago  Created                            container-B
```

### Expected Behavior

When removing all containers from `running-config`, there should be no lingering created instances in the system.

### Steps To Reproduce

_No response_

### Additional information

_No response_

### LINKED ISSUE #930: Container reconfiguration is broken
### Current Behavior

After a container has been launched any further reconfiguration is not applied until after reboot or a manual `container {stop, setup, start}` cycle.

### Expected Behavior

1. Setup `docker://nginx:alpine`
2. Activate

As expected, the default landing page is accessible

3. Add a content mount with a custom landing page
4. Activate

The custom landing page is expected to be seen, but the default is still shown

### Steps To Reproduce

_No response_

### Additional information

The root cause is that the download/setup phase for the container is run by `finit` as a `task`. So unless the runlevel changes (which it does not during a normal reconfig-cycle), the setup phase is not run again.

troglobit:
## Description

 - Bump Finit for another set of bug fixes, and `cleanup:script` support
 - Fix #902 by relocating removal from `confd` to Finit sysv `cleanup:script`
 - Fix #930 by switching from a Finit `task` (runs once per runlevel) to a `pre:script`

Also, a new container test added, to verify volumes are not lost on container upgrade.  However, upgrading is only done to the same OCI image that the container was initially created from.

As an added bonus, the `container` helper script now supports running commands inside a running Docker container.  This works even over a remote SSH connection.

## Checklist

- [X] Bugfix
  - [X] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes


troglobit:
@wkz verified on cn:styx target, no lingering containers anymore.

package/finit/finit.mk
```diff
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
-FINIT_VERSION = 4.9
+FINIT_VERSION = 4.10-rc1
```
mattiaswal, line 7: rc1, sounds dangerous.

mattiaswal (approved):
LGTM, but i think we not should run an RC of finit.

package/finit/finit.mk
troglobit: Current 4.9 plus patch backport is a lot worse.

src/confd/src/infix-containers.c
```diff
@@ -206,24 +205,18 @@ static int add(const char *name, struct lyd_node *cif)
...
-
-	/* Remove container setup script */
-	snprintf(fn, sizeof(fn), "%s/%s.sh", _PATH_CONT, name);
-	erase(fn);
-
-	/* Stop and schedule for deletion */
-	systemf("initctl -bnq stop container:%s", name);
-	writesf(name, "a", "%s", _PATH_INBOX);
+	erasef("%s/%s.sh", _PATH_CONT, name);
+	systemf("initctl -bnq disable container@%s.conf", name);
```
wkz, line 222: Is there any risk of a race between `erasef()` wiping out files that podman still expects to be available until the container has been stopped? I.e., should these two lines be reversed? Letting `podman` stop the container before cleaning up the filesystem.

Do we expect that `erasef()` will do any work here? If so, could/should it be moved to the cleanup hook of the service?

src/confd/src/infix-containers.c
```diff
@@ -54,9 +53,12 @@ static int add(const char *name, struct lyd_node *cif)
 	 * command ensures any already running container is stopped and
 	 * deleted so that it releases all claimed resources.
 	 */
+	image = lydx_get_cattr(cif, "image");
 	fprintf(fp, "#!/bin/sh\n"
+		"# meta-name: %s\n"
+		"# meta-image: %s\n"
```
wkz, line 59:
```suggestion
		"export CONTAINER_NAME=\"%s\"\n"
		"export CONTAINER_IMAGE=\"%s\"\n"
```
How about passing this info in the environment?

test/case/infix_containers/container_phys/test.py
```diff
@@ -74,7 +72,26 @@
...
+                            "name": "index.html",
+                            "content": f"{data}",
+                            "target": "/var/www/index.html"
+                        }]
+                    }]
+                }
+            })
+
+            until(lambda: url.nscheck(ns, MESG), attempts=10)
+
```
wkz, line 96:
```suggestion
        with test.step("Add a content mount, overriding index.html"):
            # Regression test for issue #930
            data = to_binary(BODY)

            target.put_config_dict("infix-containers", {
                "containers": {
                    "container": [{
                        "name": f"{NAME}",
                        "mount": [{
                            "name": "index.html",
                            "content": f"{data}",
                            "target": "/var/www/index.html"
                        }]
                    }]
                }
            })

        with test.step("Verify that index.html has been updated with the new content"):
            until(lambda: url.nscheck(ns, MESG), attempts=10)

```

From the perspective of someone trying to reproduce this test, I think it is better to break this up into two steps, and to provide some hint about what is being done.

wkz (changes requested):
Mega job! :fire:

board/common/rootfs/usr/sbin/container
```diff
@@ -109,33 +109,34 @@ EOF
...
     # Supported transports for load and create
-    case "$image" in
+    case "$uri" in
 	oci:*)			# Unpacked OCI image
-	    file=${image#oci:}
+	    file=${uri#oci:}
 	    ;;
 	oci-archive:*)		# Packed OCI image, .tar or .tar.gz format
-	    file=${image#oci-archive:}
+	    file=${uri#oci-archive:}
```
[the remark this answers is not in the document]
troglobit: Good point, and nice catch there hawkeye!  I'll update the shebang to avoid any future mishaps.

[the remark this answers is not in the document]
troglobit: > OT: Are we leaving boot-time performance on the table by not using `/bin/ash` as `/bin/sh`? Like how Debian does with `dash(1)`?

Possibly, worth exploring I think!

src/confd/src/infix-containers.c
troglobit: Good idea, but does not work in this case since it's not this script that needs the info.  Instead it's for the `container upgrade foo` command, which goes dumpster diving for the container image.

src/confd/src/infix-containers.c
troglobit: No risk of that, the `container` wrapper script and any container create/recreate/upgrade script are independent.

Yes, I think `erasef()` fits here because we create it here, in the function just prior.  

But good points nonetheless, I'll add a comment to the code about their relationship.  Because I'll likely forget myself soon :laughing:

test/case/infix_containers/container_phys/test.py
troglobit: Good point!

wkz (approved):
:sparkles: !

PR #957, opened 2025-02-23
Misc

### LINKED ISSUE #923: Skip implicit default route for IPv4 autoconf
_This issue started out as a proposal to make the "implicit default route for IPv4 autoconf" to be configurable, but [CCB](https://github.com/kernelkit/infix/issues/923#issuecomment-2642859591) decided to remove this implicit route completely. Motivation: It can be added by explicitly by those who want it. 
The description below describes the original idea._ 

### Description

It would be good to be able to disable the implicit default route which Infix currently sets up on interfaces with IPv4 autoconfiguration.
_Perhaps_ by an additional `no default-route` setting in the autoconf context 

```
admin@infix-00-00-00:/config/interface/e1/ipv4/autoconf/> help
  enabled                           Use a ZeroConf/IPv4LL agent to retrieve an 169.254/16 address.
  request-address                   Try to acquire the specified IP address, if available.
admin@infix-00-00-00:/config/interface/e1/ipv4/autoconf/> 
```

### Motivation

When configuring IPv4 autoconfiguration (169.254.x.x/16, also known as zeroconf) for an interface, Infix implicitly adds a default route through that interface. This default route has lowest possible routing PREFerence (254), so any default route learnt elsewhere will override.

The purpose of this default route is to facilitate easy setup with autoconf, even on links where there are units with only routeable addresses. See man page of [avahi-autoipd](https://www.systutorials.com/docs/linux/man/8-avahi-autoipd/ ) for the following motivation/example of added default route
```
To allow communication between hosts that have only an IPv4LL address assigned and hosts that only
have a routable IP address assigned you may add the following routes to both network configurations:

route add -net 169.254.0.0 netmask 255.255.0.0 dev eth0 metric 99

route add default dev eth0 metric 99
```

Although this implicit default route seems like a good default, it would be good if it can be disabled. In OSPF, you can make a router inject a default route into OSPF, either `always` or only when the router itself has an active default route.
The latter is done by `set default-route-advertise enabled` in OSPF context, and is good to use to avoid routing to a dead end.
This OSPF use case would be incompatible with use of IPv4 autoconfiguration on the same router, unless we can disable the implicit default route.

### Additional Information

A suitable test case would be 
- to let router A inject default route into OSPF, while having some interface with IPv4 autoconfig enabled
- to let router B see if it gets a default route via OSPF from router A

Without disabling the implicit default route, B would get a default route from A via OSPF.
If A disables the implicit default route, it will no longer advertise it and B would not get a default route.

### General Information

### LINKED ISSUE #956: Weird error while `copy factory-config running-config`
### Current Behavior

```
admin@R2:/> copy factory-config running-config 
Error: setting owner (null) on admin: Bad address
admin@R2:/> 
```

### Expected Behavior

```
admin@R2:/> copy factory-config running-config 
admin@R2:/> 
```

### Steps To Reproduce

This was reproduced with the `lingering-containers` branch on cn:styx after having run the ci:main build tests.

### Additional information

_No response_

troglobit:
## Description

 - Fix #923: drop default route from IPv4 autoconf
 - Fix #956: don't `chown()` on copy targets like `running-config`

## Checklist

- [X] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes


troglobit:
Starting with a DRAFT PR to test the waters on this one. Two *very* basic fixes that I've reset 25.02 target on now that they're fixed.  I foresee no downsides to merge.

Will do a follow-up push with updated ChangeLog as soon as the tests pass on this first push.

rical (approved):
LGTM :+1:

PR #979, opened 2025-03-09
Fix annoying "permission denied" when saving running-config

### LINKED ISSUE #977: Permission error message when 'copy running-config startup-config'
### Current Behavior

Bug seen on infix-v25.02.0-rc1.
Perhaps related to #956 

```
admin@infix-00-00-00:/> copy running-config startup-config 
Error: setting owner admin on /cfg/startup-config.cfg: Operation not permitted
admin@infix-00-00-00:/> 
```

Perhaps just a cosmetic error. The config file seems to be saved to startup.

### Expected Behavior

```
admin@infix-00-00-00:/> copy running-config startup-config 
admin@infix-00-00-00:/> 
```

### Steps To Reproduce

You can download infix-25.02.0-rc1 or the zip of a corresponding commit https://github.com/kernelkit/infix/actions/runs/13657114075
Example below is after starting QEMU (qemu.sh)

```
admin@infix-00-00-00:/> copy running-config startup-config 
Error: setting owner admin on /cfg/startup-config.cfg: Operation not permitted
admin@infix-00-00-00:/> 
```

### Additional information

_No response_

troglobit:
## Description

Fixes #977 

## Checklist

- [X] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes

### DIFF
--- a/src/bin/copy.c
+++ b/src/bin/copy.c
@@ -3,6 +3,7 @@
 
 #include <getopt.h>
 #include <pwd.h>
+#include <grp.h>
 #include <stdio.h>
 #include <stdarg.h>
 
@@ -64,6 +65,9 @@ end:
 	}
 }
 
+/*
+ * Current system user, same as sysrepo user
+ */
 static char *getuser(void)
 {
 	const struct passwd *pw;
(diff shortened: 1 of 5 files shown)


src/bin/copy.c
```diff
@@ -79,18 +83,87 @@ static char *getuser(void)
 	return pw->pw_name;
 }
 
+/*
+ * If UNIX user is in UNIX group of directory containing file,
+ * return gid, otherwise -1
+ *
+ * E.g., writing to /cfg/foo, where /cfg is owned by root:wheel,
+ * should result in the file being owned by $LOGNAME:wheel with
+ * 0660 perms for other users in same group.
```
wkz, line 92: If we made sure that we create `startup` with `g+s`, could we rely on the kernel to handle this logic for us?

troglobit: > If we made sure that we create `startup` with `g+s`, could we rely on the kernel to handle this logic for us?

You mean on `/cfg`, I assume. Yes, but I was worrying about bind + overlay mounts, as well as external media with vfat/exfat file systems.

src/bin/copy.c
```diff
@@ -79,18 +83,87 @@ static char *getuser(void)
...
+
+	getgrouplist(user, pw->pw_gid, NULL, &num);
+
+	groups = malloc(num * sizeof(gid_t));
+	if (!groups) {
+		perror("in_group() malloc");
+		return 0;
+	}
+
+	getgrouplist(user, pw->pw_gid, groups, &num);
```
rical, line 125: Isn't there's a theoretical race here, if the user is added to a group after the last call to getgrouplist()? If so, perhaps it's worth wrapping this in an if (..... != -1).

src/bin/copy.c
```diff
@@ -79,18 +83,87 @@ static char *getuser(void)
...
+	struct stat st;
+	gid_t *groups;
+	char *ptr;
+
+	pw = getpwnam(user);
+	if (!pw)
+		return 0;
+
+	strlcpy(dir, fn, sizeof(dir));
+	ptr = strrchr(dir, '/');
```
rical, line 108: Can this block use `dirname()` to make it easier to understand?

src/bin/copy.c
troglobit: There is, wrapping it with a `realloc()` is probably a good idea to be safer.

troglobit:
CCB: assigning to @rical to fix the comments.  Thanks!

troglobit (commented):
Nice work, looks good to me!

Unfortunately our policies prevent me from approving my own PR, so maybe @jovatn can make the formal approval?

src/bin/copy.c
```diff
@@ -85,37 +86,32 @@ static char *getuser(void)
 
 /*
  * If UNIX user is in UNIX group of directory containing file,
- * return gid, otherwise -1
+ * return 1, otherwise 0.
```
mattiaswal, line 89: the return of the function is still a gid_t should be int.

mattiaswal (changes requested):
Minor comment

src/bin/copy.c
rical: It "never" returned a gid_t, but yes your right. Fixed :-)

PR #986, opened 2025-03-19
Fix `mvpp2` TCAM memory corruption

wkz:
## Description

Protect the parser TCAM/SRAM memory, and the cached (shadow) SRAM information, from concurrent modifications.

Both the TCAM and SRAM tables are indirectly accessed by configuring an index register that selects the row to read or write to. This means that operations must be atomic in order to, e.g., avoid spreading writes across multiple rows. Since the shadow SRAM array is used to find free rows in the hardware table, it must also be protected in order to avoid TOCTOU errors where multiple cores allocate the same row.

This issue was detected in a situation where `mvpp2_set_rx_mode()` ran concurrently on two CPUs. In this particular case the `MVPP2_PE_MAC_UC_PROMISCUOUS` entry was corrupted, causing the classifier unit to drop all incoming unicast - indicated by the `rx_classifier_drops` counter.

## Checklist

- [x] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes

### DIFF
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -3,6 +3,13 @@ Change Log
 
 All notable changes to the project are documented in this file.
 
+[v24.11.2][] - 2025-03-19
+-------------------------
+
+### Fixes
+
+- Fix TCAM memory corruption in `mvpp2` Ethernet controller
+
 
 [v24.11.1][] - 2024-11-29
 -------------------------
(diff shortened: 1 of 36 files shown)


troglobit (approved):
Amazing work tracking this down!

PR #1004, opened 2025-04-01
Add new show command

rical:
## Description

This PR add a new show command which intend to be a safer replacement for the existing show command which this PR deprecates (renames to show-legacy). This new show command relies solely on operational data from sysrepo and attempts to ensure that a CLI user can't use it to break out of the CLI.

This is still a work in progress. Here's the main things that are left.
1) Move all cli-pretty code into the new "show" package and integrate it into the "show" command to remove the subprocess exec.
2) Continue to replace the legacy show  with operational data and the new command. Eventually removing show-legacy completely.

We are also contemplating flattening the CLI syntax to make it identical to the new show command. But the jury is still out on this. Example: "shell> show software primary" is today "cli> show software name primary". Perhaps "name" shall be removed here (it should really be partition and not name).

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes


troglobit (approved):
Much needed cleanup, nice work!

Only comment I have is that it could've been a part of the `src/bin/` family of tools, but that's something we can also collapse later as we see how all these helpers evolve over time.

PR #1038, opened 2025-05-02
 Doc upgrade/downgrade policy

### LINKED ISSUE #1009: Document upgrade/downgrade policy
Somewhere prominently there should be some information about the general upgrade/downgrade policy.  E.g.,

 1. The guaranteed way to successfully upgrade -- do a factory reset
 2. Syntax migration of startup-config after upgrade is supported
 3. Downgrading is *not supported*
 4. The guaranteed way to successfully downgrade, restore an backup of startup-config before calling the `upgrade` command

jovatn:
- Add section on upgrading/downgrading and boot-order to system.md
- Add or update references on upgrading from other pages
  - boot.md (update ref)
  - cli/upgrade.md (add ref)
  - management.md (add ref)

Fixes #1009

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes
  - [X] ChangeLog updated (for major changes)

### DIFF
--- a/doc/system.md
+++ b/doc/system.md
@@ -323,7 +323,362 @@ reference ID, stratum, time offsets, frequency, and root delay.
 > The system uses `chronyd` Network Time Protocol (NTP) daemon.  The
 > output shown here is best explained in the [Chrony documentation][4].
 
+## Upgrade procedures and boot order
+
+For resilience purposes, Infix maintains two software
+images referred to as the _primary_ and _secondary_ partition image.
+In addition, some bootloaders support [netbooting][6].
+
+The _boot order_ defines which image is tried first, and is listed
+with the CLI `show software` command. It also shows Infix version
+installed per partition, and which image was used when booting (`STATE
+booted`).
+
+```
+admin@example:/> show software
+BOOT ORDER
+primary secondary net
+
+NAME      STATE     VERSION                DATE
+primary   booted    v25.01.0               2025-04-25T10:15:00+00:00
+secondary inactive  v25.01.0               2025-04-25T10:07:20+00:00
+admin@example:/>
+```
+
+YANG support for upgrading Infix, inspecting and _modifying_ the
+boot-order, is defined in [infix-system-software][5].
+
+
+### Upgrading Infix
+
+Upgrading Infix is done one partition at a time. If the system has
+booted from one partition, an `upgrade` will apply to the other
+(inactive) partition.
+
+1. Download and unpack the release to install. Make the image *pkg*
+   bundle available at some suitable URL (FTP/TFTP/SFTP/HTTP/HTTPS).
+1. Assume the unit has booted the `primary` image. Then running the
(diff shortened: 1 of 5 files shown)


jovatn:
Perhaps too chatty documentation? :-)

doc/system.md
```diff
@@ -323,7 +323,362 @@ reference ID, stratum, time offsets, frequency, and root delay.
...
+1. Assume the unit has booted the `primary` image. Then running the
+   `upgrade` command installs a new image on the `secondary`
+   partition.
+1. As part of a successful upgrade, the boot-order is implictly
+   changed to boot the newly installed image.
+1. Reboot the unit.
+1. During boot, the unit may [migrate](#configuration-migration) the
+   startup configuration inline with newer configuration definitions.
+1. The unit now runs the new image. To upgrade the remaining partition
+   (`primary`), run the same upgrade command again, and reboot.
```
troglobit, line 369: Even though it works, thanks to Markdown, for enumerated lists I'd recommend[^0] using 1, 2, 3, ... instead of 1, 1, 1, ... and personally I try to avoid closing period for each item.  Iirc that was one of the points from Strunk & White ...

Also, for instruction lists like this I'd recommend going for brevity in each item and move any helpful hints to footnotes.  Example:

  1. Download and unpack the release to install, make the `.pkg` file available at some URL[^1]
  2. Assuming the device has booted from the primary partition, the `upgrade` command will upgrade the secondary partition
  3. ...

For the last item, reboot is really optional, so explain *why* one may want to reboot anyway.  I.e., to verify the upgrade.  Also, the second to last item should tie in better with the last, because of the risk of starting a device with a migrated configuration with an older version of Infix.  Maybe a warning or caution notice could be added here, possibly replacing one or two items?  Like this:

> [!CAUTION]
> During boot, the unit may [migrate](#configuration-migration) the startup configuration for any syntax changes.  It is therefore important that you make sure to upgrade the other partition as well after reboot, of course after having verified your setup.

[^0]: A trick in Emacs markdown mode is to press Alt-Enter when creating a new item in an itemized or enumerated list.
[^1]: Set up an FTP/TFTP/SFTP or HTTP/HTTPS server on the same LAN.

troglobit (approved):
Very well written, comprehensive and instructive!  I've given some very minor comments and ideas for improvements.  Nothing blocking merge, so this is approved in the state it is now.

The comments for lists apply to all lists in this PR even though I've just commented on the first.

doc/system.md
troglobit: Link to GitHub markdown extension for alerts https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

jovatn: Thanks! :-)

Issue #1149, opened 2025-09-08
Stopping containers sometimes fails


troglobit:
### Current Behavior

```
admin@infix-c0-ff-ee:/> show log
Sep  8 14:17:44 ha-kiosk dash[4609]: ^[[33mWARN^[[0m[0010] StopSignal SIGTERM failed to stop container dash in 10 seconds, resorting to SIGKILL
```

### Expected Behavior

We should probably extend the timeout

### Steps To Reproduce

Unknown atm.

### Additional information

Apparently the warning comes from podman itself, not conmon, not finit:

```
libpod/oci_conmon_common.go: logrus.Warnf("StopSignal %s failed to stop container %s in %d seconds, resorting to SIGKILL", sigName, ctr.Name(), timeout)
```

Issue #1148, opened 2025-09-08
Failing containers with a local OCI archive always waits for network connectivity


troglobit:
### Current Behavior

When a container setup fails, e.g., due to misconfiguration or bugs, the `container` helper script always enters a wait-for-network-retry loop, regardless where the container image is located (remote or local OCI archive.)

```
foo: setup failed, waiting for network changes ..."
```

### Expected Behavior

Should probably just exit and let Finit retry (and fail the service after a while) instead of wait for network changes.

Issue #1147, opened 2025-09-08
containers: large containers sometimes never start up


troglobit:
### Background

Testing with multiple (quite) large containers in the same setup on a Raspberrry Pi 4B, quad-core ARM Cortex-A72, 4 GiB RAM.

#### Symptom 1
 - Large containers hog CPU at startup/pull
 - Severely affects overall boot/startup time

#### Symptom 2
 - Sometimes the pulled image fails to load into `podman`
 - Manually loading image works

#### Symptom 3
 - Retries forever even if network is up
 - Manually pulling image works

### Expected Behavior

Regardless of their size, containers should start up unless the system is out of RAM or storage.

**Observations:**

 - Why do images have to be pulled every time at boot?
 - Why can't Infix cache the images between reboots?

### Steps To Reproduce

Not at time time.

### Additional information

Requiring network connection on every boot, due to the issue found in #1098, is not reason enough to wipe all images.  Hence, the solution in #1098 _needs_ to be reevaluated!

Issue #1146, opened 2025-09-08
containers: Allow to set a longer name that does not work


mattiaswal:
### Current Behavior

```
Sep  8 09:01:22 ha-kiosk container[1]: Error: setup: /run/containers/test-system-sta.sh does not exist or is not executable.
```
The name i set to the container is test-system-status
The YANG model an the system should have the same vision of the max length. (and i think it should be longer that the current used in the system)

### Expected Behavior

```
admin@infix-c0-ff-ee:/> show log
May 15 07:21:02 infix-c0-ff-ee container[3192]: Successfully created container test from curios-httpd-v24.03.0
- (press h for help or q to quit)
```

PR #1151, opened 2025-09-09
Container Lifecycle and Upgrade Improvements

troglobit:
## Description

This PR introduces significant improvements to container lifecycle management, focusing on optimization, safety, and ease of upgrades.  A long-missing container upgrade test has also been added to verify both the functionality and that old images are removed.  Not just any old downloaded OCI archives but also old/unused images loaded into the container store.

#### Container Lifecycle Optimization

  - Containers now persist across reboots unless the image or configuration changes
  - Uses checksum comparison (image archive + config) to detect when recreation is needed
  - Dramatically reduces boot time by avoiding unnecessary container recreation
  - Remote images only retry pull on network changes, not every 60s, issue
    #1146

#### Container Removal and Cleanup

  - Implements by-ID container removal to handle replacement scenarios safely, issue
    #1147 
  - When a container is removed from configuration, saves its ID for deferred cleanup
  - Finit cleanup script removes containers by immutable ID, preventing race conditions
  - Surgical cleanup: only removes images when no other containers reference them
  - Automatically cleans up OCI archives and `.sha256` sidecar files when images are removed
  - Removes old image when upgrading at startup, issue
    #1148
  - Volumes are NOT auto-removed to prevent accidental data loss (documented in user guide)

#### Container Upgrade Support

  - New upgrade RPC action for upgrading containers with mutable tags (`:latest`, `:edge`)
  - Works with both registry images and local/remote OCI archives
  - Upgrade command fixes: `admin@example:/> container upgrade NAME`

#### Documentation

  - New section explaining mutable vs immutable image tags
  - Best practices for production use (prefer immutable tags)
  - Complete rewrite of container upgrade section
  - Accurate description of container lifecycle behavior
  - Volume management documentation updates

#### Test

  - Removed redundant `container_` prefix from test directories
  - Added `V=1` support for `make test-spec` to enable debug mode
  - New container upgrade test case, issue
    #624 

#### Technical Details

  - Container optimization uses meta-sha256 and config-sha256 labels (internal)
  - Archive checksums stored in `.sha256` sidecar files in `/var/lib/containers/oci/`
  - Cleanup directory: `/var/lib/containers/cleanup/<name>/<container-id>`, issue
    #1147 
  - Increased `podman stop` timeout to prevent abrupt terminations, issue
    #1149
  - Uses nice for `podman load/create` operations to reduce system impact (at boot mostly)

## Checklist

- [x] Bugfix
  - [x] Regression tests
- [x] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [x] Test changes
- [x] Build related changes
  - [x] ChangeLog updated (for major changes)

### DIFF
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -13,12 +13,40 @@ All notable changes to the project are documented in this file.
 - Extend NETCONF and RESTCONF scripting documentation with operational
   data examples, discovery patterns, and common workflow examples, issue #1156
 - Initial support for a zone-based firewall, based on `firewalld`, issue #448
+- Add `validate` option to CLI `copy` command.  This can be used before doing a
+  restore of a backup, or when having edited configuration files manually. With
+  the validate flag (`-n` from the shell) the file is only loaded and validated
+  against the YANG models, it is *not* rolled in if validation is successful.
+  Example: `copy /media/backup/old.cfg running-config validate`, issue #373
 - Automatically expand `/var` partition on SD card at first boot on RPi
+- New `upgrade` RPC (action) for containers using images with mutable tags
+- Optimize startup of preexisting containers by adding metadata to track all
+  OCI archives loaded into container store, and all container configurations
+  used to create container instances.  Instances are now only recreated when
+  metadata from an existing instance does not match either the configuration
+  or the image — because of configuration changes or image upgrades
+- Updated container documentation on volumes, image tags, and image upgrade
 
 ### Fixes
 
+- Fix #981: copying any file, including `running-config`, to the persistent
+  back-end store for `startup-config`, does not take
+- Fix #1146: Possible to set longer containers names than the system supports.
+  Root cause, a limit of 15 characters implicitly imposed by the service mgmt
+  daemon, Finit.  The length has not been increased to 64 characters (min: 2)
+  and the YANG model now properly warns if the name is outside of these limits
+- Fix #1147: Use container metadata to clean up lingering old container images
+  instead of using the too broad `podman image prune -af` command
+- Fix #1148: Only retry container instance create on remote images
+- Fix #1149: Increase `podman stop` timeout, from 10 to 30 seconds, needed with
+  bigger containers on heavily loaded systems
 - Fix #1194: CLI `text-editor` command does not do proper input sanitation
 - Fix #1197: RPi4 no longer boots after BPi-R3 merge, introduced in v25.09
+- Upgrade fixes for containers with mutable images, e.g., `:latest`.  Infix now
+  always tries to fetch a new version of the OCI archive, for remote images,
+  regardless of the transport.  After upgrade the old image is pruned
+- Fix #1203: copying any file, including `startup-config`, to `running-config`
+  does not take
 
 [v25.09.0][] - 2025-09-30
 -------------------------


board/common/rootfs/usr/sbin/container
```diff
@@ -295,6 +307,20 @@ create()
 	args="$args --network=none"
     fi
 
+    # Add optimization labels for meta-sha256 and config checksum
+    script="/run/containers/${name}.sh"
+    if [ -f "$script" ]; then
+	# Extract meta-sha256 from script if present
```
rical: Indentation looks off.

src/confd/src/infix-containers.c
```diff
@@ -302,9 +303,34 @@ static int add(const char *name, struct lyd_node *cif)
  */
 static int del(const char *name)
 {
+	char prune_dir[sizeof(_PATH_CLEAN) + strlen(name) + 3];
```
rical: +3? Just make it PATH_MAX? Will +3 even fit `.sh` and a NULL?

src/confd/src/infix-containers.c
```diff
@@ -302,9 +303,34 @@ static int add(const char *name, struct lyd_node *cif)
...
+	if (!pp) {
+		/* Nothing to do, if we can't get the Id we cannot safely remove anything */
+		ERROR("Cannot find any container instance named '%s' to delete", name);
+		rmdir(prune_dir);
+		return SR_ERR_OK;
+	}
+
+	if (fgets(buf, sizeof(buf), pp)) {
+		chomp(buf);
+		if (strlen(buf) > 2)
```
rical: `> 2` ?

rical (commented):
Nice work!

I left some comments. Mostly cosmetic code stuff and some theoretical security issues. Nothing that you really need to address if you don't want to.

Also. I don't understand why the indentation is totally broken in the `container` script. Most lines contains a mix of tabs and spaces. It's hard to read on GH and my editor is contemplating `kill -9 $$` ;)

src/bin/copy.c
```diff
@@ -320,9 +366,29 @@ static int copy(const char *src, const char *dst, const char *remote_user)
...
+						rc = replace_running(conn, sess, fn, timeout);
+						if (!rc)
+							set_owner(dstds->path, user);
+					}
+					sr_disconnect(conn);
+				}
+			} else {
+				rc = systemf("sysrepocfg -d %s -I%s -f json", dstds->sysrepocfg, fn);
+				if (rc)
+					fprintf(stderr, ERRMSG "failed loading %s from %s", dst, src);
```
rical: Missing `\n`

src/bin/util.c
```diff
@@ -81,38 +82,84 @@ static const char *basenm(const char *fn)
...
-	}
-
-	/* Files in /cfg must end in .cfg */
-	if (!strncmp(fn, "/cfg/", 5)) {
+	if (sanitize) {
+		char testpath[PATH_MAX];
+		/* CLI mode - restricted and helpful */
+
+		/* Block path traversal attacks */
+		if (strstr(fn, "../"))
```
rical: Just a heads up. If this path is later used together with something, like `("%s/%s", fn, name)` an attacker could place `..` in `fn` like `/etc/sysrepo/..` and then set `name` to `shadow`. This would bypass this check and result in `/etc/sysrepo/../shadow`, or if he/she can control the name even `(%s%s", fn, name)` would be vulnerable as name could be `/shadow`.

Not that I believe it's a problem here.

src/bin/util.c
```diff
@@ -81,38 +82,84 @@ static const char *basenm(const char *fn)
 	return ptr;
 }
 
-char *cfg_adjust(const char *fn, const char *tmpl, char *buf, size_t len)
+char *cfg_adjust(const char *fn, const char *tmpl, char *buf, size_t len, int sanitize)
```
rical: As we attempt to sanitize, do we need to care about symlinks that points outside the scope? I don't really have the context here.

troglobit: Awesome, thanks!  I ended up with a whitelist of absolute paths (/cfg, /media, $HOME), that I check against twice, the second time after having called `realpath()` on the sanitized filename.

test/case/infix_containers/container_phys/topology.dot
```diff
@@ -1 +0,0 @@
-../container_bridge/topology.dot
```
mattiaswal: Why this?

troglobit:
> > > Also. I don't understand why the indentation is totally broken in the `container` script. Most lines contains a mix of tabs and spaces. It's hard to read on GH and my editor is contemplating `kill -9 $$` ;)
> > Well, Emacs default indentation for shell scripts, as most of us use, have a four space base indent, but replace leading 8-char spaces with tabs. I think one of the main reasons for your review issues is that GitHub has changed (!) the tab width since 2021 from 8 to 4 bytes, but this is possible to change in your https://github.com/settings/appearance settings
> I strongly disagree towards mixing tabs and spaces for indentation. Personally I prefer spaces, which most documentation suggest (google et al). But growing up in #bash on IRC getting yelled at by GreyCat, I can live with tabs too.

I'm sorry this turned into a tabs-vs-spaces discussion, that was not my intention.  I was just stating a fact that most of our scripts have this mix of tabs vs spaces because of us Emacs users.

> If you look at the file in the current state you will see that there's a mix of everything. Some lines are pure space (8+), some are pure tabs and some are a mix of tabs and spaces.

Yes, and I'd rather not reformat the entire file at this point because that would make `git blame` a lot more difficult.

mattiaswal (changes requested):

test/case/infix_containers/container_phys/topology.dot
troglobit: I did not actively go in and remove the newline, if that's what you're asking?  I've renamed/simplified the test directory names and both these files are symlinks to the container bridge test's `topology.dot` file.

troglobit: Aha, you thought I'd removed them?  Just GIT having a bad day I think, didn't manage to track the rename:

```
 test/case/infix_containers/container_phys/topology.dot                               |  1 -
 test/case/infix_containers/container_veth/topology.dot                               |  1 -
 test/case/infix_containers/phys/topology.dot                                         |  1 +
 test/case/infix_containers/veth/topology.dot                                         |  1 +
```

Issue #1080, opened 2025-07-10
Error message in log from rauc


mattiaswal:
### Current Behavior

When initiate an upgrade, following is in the log:

Jul 10 06:10:16 infix-00-00-00 rauc[3587]: Using deprecated 'Install' D-Bus Method (replaced by 'InstallBundle')

### Expected Behavior

No error.

PR #1164, opened 2025-09-27
Rauc integration fixes

troglobit:
## Description

 - #1080 
 - Two new optional arguments to `install-bundle` RPC
   - `verify-certificate` — weather to verify bundle certificate at install, default: `True`
   - `verify-compatible` — weather to verify if bundle is compatible, default: `True`
 - Drop redirect of rauc console events (ANSI color coded) in favor of dedicated JSON event log

## Checklist

- [x] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes

### DIFF
--- a/src/confd/src/infix-system-software.c
+++ b/src/confd/src/infix-system-software.c
@@ -31,6 +31,7 @@ static int infix_system_sw_install(sr_session_ctx_t *session, uint32_t sub_id,
 	sr_error_t srerr = SR_ERR_OK;
 	GError *raucerr = NULL;
 	RaucInstaller *rauc;
+	GVariant *args;
 
 	DEBUG("url:%s", url);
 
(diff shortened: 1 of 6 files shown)


src/confd/src/infix-system-software.c
```diff
@@ -27,20 +27,55 @@ static int infix_system_sw_install(sr_session_ctx_t *session, uint32_t sub_id,
 				   unsigned request_id, sr_val_t **output,
 				   size_t *output_cnt, void *priv)
 {
-	char *url = input->data.string_val;
+	char *url = NULL;
+	gboolean verify_certificate = TRUE;  /* default value */
+	gboolean verify_compatible = TRUE;   /* default value */
```
wkz, line 32: AFAIK, beyond these two guardrails, there are no further sanity checks in the install scripts or anything like that. Right? If you feed it a bundle with `LOLCAT.GIF` renamed to `rootfs.squashfs` we will happily flash that to primary.

When, in a non-development situation, do we foresee this not ending in tears for the user? I.e., when can we ignore a certificate mismatch and not have the bootloader just refuse to boot the image after a reboot anyway?

If the use-case is going between spins and/or vanilla, or vice versa, then is that not rare enough (and likely in a development setting) that we can upgrade via `rauc(1)` over SSH or something?

troglobit: Sure, I was just going off my suggestion in #1080 that we seemingly approved in the CCB. But you're right, these can be dropped since they are not really useful to non-devs.

troglobit: Patch dropped.

wkz: Alright. Just for the record though: I am happy to have my mind changed on this. I just wanted to understand the use-case.

PR #1166, opened 2025-09-28
Mix of test framework fixes and improvements

troglobit:
## Description

A mix of test framework fixes and improvements extracted from the firewall branch for separate review.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes
- [x] Other (please describe): infamy

### DIFF
--- a/test/infamy/restconf.py
+++ b/test/infamy/restconf.py
@@ -26,16 +26,14 @@ class Location:
 
 def xpath_to_uri(xpath, extra=None):
     """Convert xpath to HTTP URI"""
-    # If the xpath has a
     pattern = r'\[(.*?)=["\'](.*?)["\']\]'
     matches = re.findall(pattern, xpath)
 
+    uri_path = xpath
     if matches:
         for key, value in matches:
             # replace [key=value] with =value
-            uri_path = re.sub(rf'\[{re.escape(key)}=["\']{re.escape(value)}["\']\]', f'={value}', xpath)
-    else:
-        uri_path = xpath
+            uri_path = re.sub(rf'\[{re.escape(key)}=["\']{re.escape(value)}["\']\]', f'={value}', uri_path)
 
     # Append extra if provided
     if extra is not None:
--- a/test/infamy/tap.py
+++ b/test/infamy/tap.py
@@ -30,7 +30,7 @@ class Test:
         self.out.flush()
         return self
 
-    def __exit__(self, _, e, __):
+    def __exit__(self, t, e, tb):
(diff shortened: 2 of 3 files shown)


troglobit:
We got a "cut off" job here due to duplicate build, but the Qemu test results are here https://github.com/kernelkit/infix/actions/runs/18068123992

test/infamy/restconf.py
```diff
@@ -26,16 +26,19 @@ class Location:
...
-            # replace [key=value] with =value
-            uri_path = re.sub(rf'\[{re.escape(key)}=["\']{re.escape(value)}["\']\]', f'={value}', xpath)
-    else:
-        uri_path = xpath
+            if key == '.':
+                # For [.='value'], replace with =value
+                uri_path = re.sub(rf'\[\.=["\']{re.escape(value)}["\']\]', f'={value}', uri_path)
+            else:
+                # For [key='value'], replace with =value
+                uri_path = re.sub(rf'\[{re.escape(key)}=["\']{re.escape(value)}["\']\]', f'={value}', uri_path)
```
wkz, line 41:
```suggestion
    uri_path = re.sub(r"\[(.|\w+)='(\w+)'\]", r"=\2", xpath)
```

Not sure if I understand everything that is going on here, but is there any case that the current implementation handles, that would not covered by this suggestion?

test/infamy/tap.py
```diff
@@ -95,7 +99,10 @@ def succeed(self):
     def skip(self):
         raise TestSkip()
 
-    def fail(self):
+    def fail(self, message=None):
+        if message:
+            self.out.write(f"# {message}\n")
+            self.out.flush()
```
wkz, line 105: Are there any cases where we could not use `assert`s instead? I.e., was the original sin me introducing `fail()` in the first place?

wkz (changes requested):
We don't necessarily get rid of `test.fail()` now. It was just something that struck me as I read the log message.

test/infamy/restconf.py
troglobit: Nice point, I got lost in regexps trying to fix this one.  The actual bug was that we used xpath for each key,value match, instead of the converted value from any previous call in the for loop.

test/infamy/tap.py
troglobit: Well, even though assert probably is more pythonic, I still think success/fail constitute a clear API.  I'd vote to keep as-is for now, with this addition.

PR #1316, opened 2025-12-09
Add operational data journal with retention policy

rical:
## Description

Dump operational datastore to timestamped JSON snapshots every 5 minutes (in /var/lib/statd/). The operational.json symlink always points to the latest snapshot.

Implement hierarchical retention policy that keeps the first snapshot of each time period (hour/day/week/month/year), providing fine-grained recent history while preventing unbounded disk usage.

This will allow us to plot / track how the system state evolves as well as give us somewhat fine-grained info in the case of an event, such as a crash.

Add unit test simulating months of snapshots to verify retention behavior using a statd stub that only runs the retention code locally (unit test)

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [x] Feature
  - [x] Regression tests added?
  - [ ] Documentation added?
- [x] Test changes
  - [x] New Unit Test
- [ ] Build related changes

### DIFF
--- a/src/statd/journal.c
+++ b/src/statd/journal.c
@@ -11,6 +11,7 @@
 #include <sys/stat.h>
 #include <pthread.h>
 #include <dirent.h>
+#include <zlib.h>
 
 #include <srx/common.h>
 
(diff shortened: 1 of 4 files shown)


src/statd/journal.c
```diff
@@ -0,0 +1,198 @@
...
+	int ret;
+
+	now = time(NULL);
+	get_timestamp_filename(timestamp_file, sizeof(timestamp_file), now);
+	snprintf(timestamp_path, sizeof(timestamp_path), "%s/%s",
+		 JOURNAL_DIR, timestamp_file);
+
+	/* Write timestamped snapshot directly */
+	ret = lyd_print_path(timestamp_path, tree, LYD_JSON,
+			     LYD_PRINT_WITHSIBLINGS);
```
wkz: Did some quick testing on a 28-port system with more or less an empty config:

```
root@test-05-25-f8:/tmp$ sysrepocfg -X -d operational -f json >oper.json
root@test-05-25-f8:/tmp$ ll -h oper.json
-rw-r--r--    1 root     root      237.9K Dec  9 22:36 oper.json
root@test-05-25-f8:/tmp$ gzip oper.json
root@test-05-25-f8:/tmp$ ll -h oper.json.gz
-rw-r--r--    1 root     root       19.6K Dec  9 22:37 oper.json.gz
```
Should we pass this through `libz` before going to disk? I'd hate for this storage to become an issue on smaller systems or as the amount of generated data grows. With compression, disk usage after one year should go from 14MB down to 1MB with today's feature-set. That should set us up to not have to think about this for a _long_ time.

wkz (changes requested):
Very nice! :fire:

PR #1384, opened 2026-01-30
Upgrade FRR to 10.5.x

### LINKED ISSUE #1029: FRR: Adapt to the changes in FRR 10.x
In FRR 10 they changed how it is supposed to use FRR. Now you can not have multiple configuration files (one for ospfd, one for staticd and so on). They already changed this in 9.1 but it was possible to patch away the new behavior, from 10 it is not possible/desired.

This does not mean that we need to upgrade to 10.x just that it possible.

The 9.1 branch in FRR is more or less abandoned.

This is a rather big job, not desired to be forced to do this because we **must** upgrade FRR due to bugs/CVEs.

mattiaswal:
Upgrade FRR to 10.5.1

This includes adding a netd-daemon since FRR has gone towards
all daemons should be configured from mgmtd using YANG,
netd uses gRPC to reconfigure in mgmtd.

The netd implmentation also support to to disable FRR buildtime and
still have route management using kernel routes. This is feature should
 be treated as pretty untested and exprimental.

This fix #1029

## Checklist

- [x] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes


board/common/rootfs/usr/share/udhcpc/default.script
```diff
@@ -102,7 +119,8 @@ set_dhcp_routes()
     cmp -s "$NAME" "$NEXT" && return
     mv "$NEXT" "$NAME"
 
-    initctl -nbq restart staticd
+    killall -SIGHUP netd
+    killall staticd
```
troglobit: I suggest `initctl reload netd` instead, that way if it is for some reason is disabled or in a wait state, Finit will handle that too.

Also, if the intention is to restart `staticd`, please use `initctl restart staticd` instead, otherwise the respawn counter in Finit (used to track malfunctioning services) is incremented and might cause the Finit from marking staticd as faulty.

troglobit (changes requested):
Wow, this was an amazing achievement! 🤯 It will be a huge step forward and also very useful for our under-powered boards that don't need a full blown Frr just to set static routes.

Issue #1361, opened 2026-01-21
ntp: client_stratum_selection: Fails to select lowest stratum sometimes


mattiaswal:
### Current Behavior

```
TOPOLOGY=/etc/infamy-styx.dot  TEST_MODE=host make test-sh

export PYTHONHASHSEED=3124318344
while [ $? -eq 0 ]; do ./9pm/9pm.py -a ./case/ntp/client_stratum_selection/test.py  ; done

2026-01-21 08:26:04 7..7

o Execution
`-- o 0001 test.py
9PM - Simplicity is the ultimate sophistication (0cea38dc1e)

Testing Infix (3191306a5af6)

Starting test 0001 test.py (../case/ntp/client_stratum_selection/test.py)
2026-01-21 08:26:04 # Starting (2026-01-21 08:26:04)
2026-01-21 08:26:04 # host:    monza
2026-01-21 08:26:04 #     mgmt1:   sw1p4
2026-01-21 08:26:04 #     srv1:    sw1p8
2026-01-21 08:26:04 #     mgmt2:   sw1p2
2026-01-21 08:26:04 # srv2:    styx4
2026-01-21 08:26:04 #     mgmt:    e28
2026-01-21 08:26:04 #     swp1:    e27
2026-01-21 08:26:04 #     swp2:    e17
2026-01-21 08:26:04 # client:  styx2
2026-01-21 08:26:04 #     eth0:    e17
2026-01-21 08:26:04 #     mgmt:    e28
2026-01-21 08:26:04 #
2026-01-21 08:26:04 # Waiting for DUTs to become reachable...
2026-01-21 08:26:04 # Probing styx4 on port sw1p4 for IPv6LL mgmt address ...
2026-01-21 08:26:04 # Testing using RESTCONF
2026-01-21 08:26:05 # YANG models downloaded.
2026-01-21 08:26:07 # Waiting for DUTs to become reachable...
2026-01-21 08:26:08 # Probing styx2 on port sw1p2 for IPv6LL mgmt address ...
2026-01-21 08:26:08 # Testing using RESTCONF
2026-01-21 08:26:08 # YANG models downloaded.
2026-01-21 08:26:10 ok 1 - Set up topology and attach to devices
2026-01-21 08:26:12 ok 2 - Configure srv2 to sync from srv1 and serve with higher stratum
2026-01-21 08:26:17 ok 3 - Wait for srv2 to sync from srv1
[...]

Issue #892, opened 2025-01-15
Worrying rousette warnings in log


troglobit:
Suggest patching these two logs that might look a bit worrying to end users:

```
Jan 14 05:42:12 ix-00-00-00 rousette[3439]: [2025-01-14 05:42:12.277] [rousette] [info] NACM config validation: Anonymous user access disabled 
Jan 14 05:42:12 ix-00-00-00 rousette[3439]: [2025-01-14 05:42:12.278] [rousette] [warning] Telemetry disabled. No CzechLight YANG modules found. 
```

mattiaswal:
Agree!

mattiaswal:
I would say that the first "info" is more of a security issue that needs to be patched away in rousette (and upstreamed):
From the Readme of rousette, annonymous access is enabled when:
```
1. The first entry of `rule-list` list must be configured for `ANONYMOUS_USER_GROUP`.
2. All the rules except the last one in this rule-list entry must enable only "read" access operation.
3. The last rule in the first rule-set must be a wildcard rule that disables all operations over all modules.

The anonymous user access is disabled whenever these rules are not met.

```

mattiaswal:
For anonymous we should add a switch to rousette that prefered enable anonymous mode, and test if they accept it.

PR #1412, opened 2026-02-22
Optimize boot time and consolidate confd into a single daemon                                                             

troglobit:
## Description

This PR replaces `sysrepo-plugind` + `bootstrap` + `load` scripts with a single `confd` that handles config generation, datastore init, config loading, and plugin management. The daemon uses a `libev` event loop with `SR_SUBSCR_NO_THREAD` instead of ~30 per-subscription sysrepo threads.

The new initviz package was used for boot time visualization.

Initial work on this branch speed up `sysctl-sync-ip-conf` and `mnt` scripts, as well as move `hostname.d` setup to build-time instead of runtime.  To optimize the bootstrap process and free up CPU time on single-core systems, the start of many services have been postponed to either runlevel 2, or after `confd` bootstrap, like  `dbus`, `dnsmasq`, and `statd`.  To reduce overhead, we also drop unnecessary logger processes for services that already log to syslog.  Even BusyBox has been inspected, we now enable NOEXEC/NOFORK applets.

Other interesting changes in this PR is to drop WiFi and GPS from minimal defconfigs.  Move WiFi/firmware selects from RPi board `Config.in` to the full defconfigs so features are explicit rather than implicit via BSP.

**Additional changes on this branch:**

 - User-configurable HTTPS certificate: A new certificate node under services web lets users select any keystore certificate for the HTTPS management
  interface.  The default auto-generated self-signed cert is now stored in the keystore (migration script included for existing devices).
 - confd crash recovery: A sentinel file on tmpfs lets confd distinguish a crash-restart from a fresh boot and skip destructive bootstrap phases (sysrepo SHM wipe, factory config reinstall) on restart, preserving a consistent datastore state across unexpected restarts.
 - NTP stratum-weight fix: The stratum-weight YANG leaf and its description had inverted semantics — 0.0 disables stratum weighting entirely (pure distance-based selection), not the other way around.  The client_stratum_selection test now uses 1.0, which gives each stratum level a 1-second effective advantage and guarantees deterministic stratum-based selection on local networks.  Fixes the long-standing flaky test (#1361).
 - Log noise reduction: Silenced spurious warnings from rousette (#892) and netopeer2-server (#1446).  Transient mDNS daemon restart messages in statd are now suppressed unless the mDNS service is actually enabled in config.  The "avahi" subsystem is renamed to "mdns" in all user-facing log messages.
 - CI: Fixed Coverity Scan build broken by the new avahi/mDNS dependency in statd (libavahi-client-dev was missing from the apt-get step).

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [x] Feature
  - [ ] Regression tests added?
  - [x] Documentation added?
- [x] Test changes
- [x] Build related changes

### DIFF
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -31,11 +31,24 @@ All notable changes to the project are documented in this file.
   - `no ssh known-hosts <host>` — remove a stale or changed host key entry
 - Add OSPF point-to-multipoint (P2MP) and hybrid interface type support.  This
   also includes support for setting static neighbors, issue #1426
+- Add support for user-configurable HTTPS certificate: select any certificate
+  stored in the keystore via `services web certificate <name>`.  The default
+  auto-generated self-signed certificate is now stored in the keystore rather
+  than in `/cfg/ssl` and existing certificates are auto-migrated on upgrade
 
 ### Fixes
 
+- Fix #892: suppress rousette RESTCONF server warnings from syslog
+- Fix #1387: `infix.local` now resolves to exactly one device per LAN.  Previously
+  all Infix devices claimed both `hostname.local` and `infix.local`, causing avahi
+  to append `#2`, `#3` suffixes to the shared alias on busy networks.  Assignment
+  is now first-come-first-served using standard mDNS conflict resolution
+- Fix #1416: `show firewall` command show an error when the firewall is disabled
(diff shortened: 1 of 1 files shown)


troglobit:
@mattiaswal + @wkz this PR is now finally done and ready for review.

PR #1457, opened 2026-03-29
Initial web management interface

troglobit:
## Description

First WebUI for Infix.  Uses go + htmx as middleware, custom javascript for the frontend (atm.), and RESTCONF from the middleware to 127.0.0.1.

Based on [kernelkit/webui2](https://github.com/kernelkit/webui2) by @mattiaswal 

## Screenshots

login

dashboard

wizard

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [X] Feature
  - [x] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes


board/common/rootfs/etc/nginx/restconf-access.conf
wkz, line 1: If this was `/etc/nginx/restconf-access-local.conf`, and the rootfs shipped with 

`restconf-access.conf -> restconf-access-local.conf`, 

then we would not have to duplicate the contents of this file in `confd` - potentially saving us sync issues in the future.

board/common/rootfs/etc/default/webui
```diff
@@ -0,0 +1,5 @@
+RESTCONF_URL=https://127.0.0.1/restconf
+INSECURE_TLS=1
+# Spool firmware uploads (and any other temp files) to the eMMC-backed /var/tmp
+# instead of the RAM-backed /tmp — a 160 MB .pkg upload otherwise OOM-kills us.
```
wkz:
```suggestion
# Spool firmware uploads (and any other temp files) to persistent storage
# instead of the RAM-backed /tmp to enable push-upgrades on low-memory systems.
```
Just a nit, but in general: think we should avoid referencing technologies, sizes, etc., that might quickly become outdated and confuse the reader.

package/webui/webui.conf
wkz, line 1: Lots of the information in these comments feels like Claude justifying its decisions. That is great information to have, but to me it feels like it belongs in the Git log 🤷‍♂️

src/webui/internal/auth/login.go
```diff
@@ -0,0 +1,144 @@
...
+package auth
+
+import (
+	"context"
+	"errors"
+	"html/template"
+	"log"
+	"net/http"
+
+	"github.com/kernelkit/webui/internal/handlers"
```
wkz: Should this be 
```suggestion
	"github.com/kernelkit/infix/src/webui/internal/handlers"
```
now that we're importing it? Can you set it to something shorter? `infix/webui` or something?

src/webui/internal/auth/login.go
```diff
@@ -0,0 +1,144 @@
...
+	username := r.FormValue("username")
+	password := r.FormValue("password")
+
+	if username == "" || password == "" {
+		h.renderLogin(w, r, "Username and password are required.")
+		return
+	}
+
+	// Verify credentials by making a RESTCONF call with Basic Auth.
+	err := h.RC.CheckAuth(username, password)
```
wkz, line 59: Do we have any rate-limiting in place for reaching this point? I presume this involves some non-trivial amount of work, so could be a DOS-vector?

configs/aarch64_minimal_defconfig
```diff
@@ -144,6 +144,7 @@ BR2_PACKAGE_FIREWALL=y
 BR2_PACKAGE_IITO=y
 BR2_PACKAGE_KEYACK=y
 BR2_PACKAGE_KLISH_PLUGIN_INFIX=y
+BR2_PACKAGE_LANDING=y
```
wkz, line 147: What does `landing` pull in? Do we need it in minimal?

src/webui/internal/auth/store.go
```diff
@@ -0,0 +1,156 @@
...
+// persist across restarts.
+type SessionStore struct {
+	aead cipher.AEAD
+}
+
+// NewSessionStore creates a store.  If keyFile is non-empty, the AES
+// key is read from that path (or generated and written there on first
+// run).  If keyFile is empty, a random ephemeral key is used.
+func NewSessionStore(keyFile string) (*SessionStore, error) {
+	key, err := loadOrCreateKey(keyFile)
```
wkz: It feels dangerous to me to not have this key in the regular `keystore` 🤔 

If I have a box, with an active WebUI session as the `admin` user, and I then run `copy factory-config running-config`, my session would still be valid, no? If the AES key was in the keystore, it would be regenerated like the self-signed certificates, and all previous sessions would be automatically invalidated.

wkz (changes requested):
Massive work! 🤯 So impressed! 🙏 

The actual implementation is mostly out of my comfort zone as far as reviewing goes.

There is no testing whatsoever, AFAIK. Should we have something that at least tries to login and load the dashboard?

Maybe superficial, but: the icons feel a bit all over the place. AI? Would it feel more coherent if we picked a set from flaticon or similar?

configs/aarch64_minimal_defconfig
troglobit: It's a README.md -> HTML index page when web (i.e., RESTCONF is enabled), and with this PR a different nginx.conf file, basically.

src/webui/internal/auth/store.go
troglobit: The whole thing is a bit of an anti-pattern really, it was put in there to "comfortable" auto-login users after upgrade/reboot. But I think just replacing this "master key" approach with an in-memory caching of a per-user/per-session token is the safest route with a lot less complexity as well.

User's will experience a *slight discomfort* of having to log in again after upgrade/reboot, but I think that's a reasonable compromise and similar to what other services expect after a session timeout or other operational changes to infra.

src/webui/internal/auth/login.go
troglobit: Absolutely, fixing!

package/webui/webui.conf
troglobit: Yup, I've been doing a lot of cleanup already ... :sweat_smile:

board/common/rootfs/etc/default/webui
troglobit: Good point, and nice catch, will fix!

board/common/rootfs/etc/nginx/restconf-access.conf
troglobit: Hmm, fair point, I'll look into it!

src/webui/internal/auth/login.go
troglobit: Not yet, but as it turns out nginx ships with a standard module for exactly this:
```
  # In webui.conf or webui-proxy.conf, top of file:
  limit_req_zone $binary_remote_addr zone=webui_login:10m rate=20r/m;

  # Then on the login endpoint:
  location = /login {
      limit_req zone=webui_login burst=5 nodelay;
      include /etc/nginx/webui-proxy.conf;
  }
```

Will add in the next update of this PR.  Thank you for bringing it up!

mattiaswal:
I am fine with this, much better than before, and 0 security issues find. Great work :muscle:

src/webui/internal/auth/store.go
troglobit: Fixed.

src/webui/internal/auth/login.go
troglobit: Fixed.

board/common/rootfs/etc/nginx/restconf-access.conf
troglobit: Fixed.

troglobit:
> There is no testing whatsoever, AFAIK. Should we have something that at least tries to login and load the dashboard?

@mattiaswal donated a really good test.

> Maybe superficial, but: the icons feel a bit all over the place. AI? Would it feel more coherent if we picked a set from flaticon or similar?

Icon set consolidated and now use [Lucide](https://lucide.dev/). Hope this is OK for v1.

wkz (approved):
Let's gooooooooooooO!

PR #1498, opened 2026-05-11
Add cellular modem (wwan) support

troglobit:
## Description

Add cellular modem (WWAN) support for USB-attached MBIM/QMI modems, including USB dongles, mPCIe cards, and M.2 Key-B modules.  Multiple APNs per modem, SIM PIN configuration, and NMEA routing from modem GPS to gpsd.

See the [Modem User's Guide](https://github.com/kernelkit/infix/blob/modem/doc/modem.md) for details.

Initially donated by [Avvero Pty](https://avvero.com/), the original YANG model has been significantly redesigned, split into modem and sim components under ietf-hardware, and wwan as a dedicated interface type in ietf-interfaces.  Finally, the location handling is now fully handed off to the Infix GPS subsystem (another ietf-hardware component), allowing chrony to use the modem as a time source.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [x] Feature
  - [ ] Regression tests added?
  - [x] Documentation added?
- [ ] Test changes
- [ ] Build related changes


doc/ChangeLog.md
```diff
@@ -10,21 +10,26 @@ All notable changes to the project are documented in this file.
...
 - Add Wi-Fi 802.11s mesh support, letting access points form a wireless
   backhaul between each other without cabling
 - Add band steering for dual-band access points, nudging dual-band
   clients onto the faster 5/6 GHz band
 - Add `legacy-rates` option to re-enable 802.11b rates on 2.4 GHz for
   old IoT devices (disabled by default)
+- Add cellular modem (WWAN) support for USB-attached MBIM/QMI modems,
+  including USB dongles, mPCIe cards, and M.2 Key-B modules.  Multiple
+  APNs per modem, SIM PIN configuration, and NMEA routing from modem
+  GPS to gpsd.  See the [Modem User's Guide][modem] for details
```
mattiaswal, line 23: Maybe strip the gpsd part here (implementation detail), instead point to how we handle GPSes?

mattiaswal:
The interface behavior is very similar how it works in Wi-Fi, one interface is created automatic by the kernel, but Infix general design demands confd to own all interfaces. The solution for this in Wi-Fi was:

1) keep wlan0, create virtual interfaces on top of it
  this was problematic since a lot of cards only supported ONE virtual interface (wlan0 was also virtual, guess it is the
  same in wwlan)
 2) Remove wlan0 as soon as it was detected, then create the virtual interfaces requested by the user
     This worked, since the wifi stack had a radio handle (phy0) which we could connect new wi-fi interfaces on
     
     
  In wi-fi i went for 2, now all wi-fi interfaces is the same as other interfaces, reference from other ip commands etc. 
  Is there any similar radio-reference in wwlan? Since this automatic approach i fear would lead to pain and failed  
 dagger settings.
  
  Claude suggested this to me:
  ```
  

Modern WWAN-core links (kernel ≥ 5.14, drivers like mhi_wwan_mbim, iosm, mtk_t7xx). Here the data netdev is not a
 fixed thing — it's a link you add on top of the parent wwanX device, roughly ip link add ... type wwan parentdev 
wwanX linkid N, and you can ip link del it freely. This is the case where "remove it, let confd create it later" works 
cleanly, because creation is an explicit operation anyway (normally done by ModemManager).

```

troglobit:
> The interface behavior is very similar how it works in Wi-Fi, one interface is created automatic by the kernel, but Infix general design demands confd to own all interfaces. The solution for this in Wi-Fi was:
> 
>     1. keep wlan0, create virtual interfaces on top of it
>        this was problematic since a lot of cards only supported ONE virtual interface (wlan0 was also virtual, guess it is the
>        same in wwlan)
> 
>     2. Remove wlan0 as soon as it was detected, then create the virtual interfaces requested by the user
>        This worked, since the wifi stack had a radio handle (phy0) which we could connect new wi-fi interfaces on
> 
> 
> In wi-fi i went for 2, now all wi-fi interfaces is the same as other interfaces, reference from other ip commands etc. Is there any similar radio-reference in wwlan? Since this automatic approach i fear would lead to pain and failed dagger settings.
> 
> Claude suggested this to me:
> 
> ```
> 
> 
> Modern WWAN-core links (kernel ≥ 5.14, drivers like mhi_wwan_mbim, iosm, mtk_t7xx). Here the data netdev is not a
> fixed thing — it's a link you add on top of the parent wwanX device, roughly ip link add ... type wwan parentdev 
> wwanX linkid N, and you can ip link del it freely. This is the case where "remove it, let confd create it later" works 
> cleanly, because creation is an explicit operation anyway (normally done by ModemManager).
> ```

Thank you for the review, this is definitely worth considering since so much in Infix hinges on the interface existing when everything else is unleashed.  I'll have a look at it and see what I can do!

troglobit:
Marking as draft again to look into @mattiaswal's very valid comments.

troglobit:
Back at it again, short answer: there's no `phy0` equivalent for USB modems. The `parentdev`/`linkid` API is wwan-core only (`mhi_wwan_mbim`, `iosm`, `mtk_t7xx` — all PCIe), and `cdc_mbim`/`qmi_wwan` netdevs are driver-owned `usbnet` devices that can't even be deleted.  So the WiFi delete-and-recreate approach doesn't transfer 1:1 unfortunately.  Your worry is well founded though because we ran into exactly that on the BPI-R4: the dummy grabbed `wwan0` while the EM05 modem was still bouncing through its 60s cold boot, and the real interface came up as `wwan1`.

Plan for USB modems instead:

1. The modem hardware component gets a USB `devpath` setting, configured by the user.  Modems are pluggable so we never pre-provision them in `factory-config`; support finds the `devpath` and configures the component.
2. ModemManager runs with `--filter-policy=strict` and only gets devices `confd` tags for it.
3. A udev rule generated from config renames whatever `netdev` appears under the `devpath` to the configured interface name.  Deterministic regardless of enumeration order, so the dummy placeholder goes away along with the name race.
4. Modem absent after `probe-timeout`: skip gracefully, same as WiFi.

So `confd` owns identity, naming, and the MM handover, even though the kernel still creates the `netdev`. When true PCIe (`wwan-core`) modems show up they anchor the same way, and there `confd` can create the links for real via `parentdev`/`linkid`.

**Bonus:** the same `devpath`-anchored component pattern fits USB GPS dongles, which fixes deterministic naming for those too.

PR #1502, opened 2026-05-13
Upgrade FRR to 10.5.4

mattiaswal:
Summary of Changes:

bfdd:
    - cap IPv6 echo reflection to declared length
    - account for FP offset in echo length checks
    - fix recv errno filter logic in a few places
    - tighten SBFD reflector packet sanity checks
    - gate IPv6 echo reflection on known sessions
    - tighten auth header parsing skeleton
    - validate control packet length before session lookup

bgpd:
    - Fix memleak when configuring rd
    - Validate if NHC BGPID TLV value is non-zero
    - Avoid having a dangling pointer after we free NHC attribute
    - Check if BGPID NHC TLV exists when IPv6 next-hop is link-local
    - Do not allocate NHC TLV with an extra trailer
    - migrate timers during peer_xfer_conn to fix stale route cleanup
    - honor 'no activate' for dynamic neighbors in peer-group
    - Return immediately when dynamic capability action is not valid
    - Validate BGP role capability when handling it dynamically
    - fix neighbor IP comparison for IPv6 memcmp return values
    - Don't mark nexthop as changed if a set next-hop unchanged is applied
    - Return BGP_PEER_INTERNAL when first peer's as type is set to auto
    - Update peer sort cache when remote-as auto is used and AS number changed
    - Check dynamic capability action before validating ENHE capability
    - Do not allocate stream if route-refresh capability is not received
    - Move rpki strict check to bgp_accept()
    - Fix memory leak for nhc attribute if ipv6 is link-local address
    - Fix compilation for Debian 11 when printing uint64 values
    - Return zero labels if no BOS found and it's not a withdraw label
    - Fix signed overflow in hexstr2num()
    - Check the length also when parsing ENCAP attr sub-TLVs
    - Validate prefixlen before subtracting when parsing labeled unicast NLRI
    - Reset the stream to attr_start + attribute_len when WITHDRAWN
    - Revalidate locally originated routes against RPKI changes
    - Check if prefixlen is not 0 when parsing flowspec stuff
    - Prevent len_string going negative when trying to display flowspec entries
    - fix import vrf on non existing vrf
    - fix no vrf import command
    - Free hostname for FQDN capability if the parsing goes wrong
    - Validate MP_REACH_NLRI attribute against incorrect next-hop
    - Fix dynamic FQDN capability handling
    - Check if the remaining length for subtracting TLV length is enough
    - Fix the end pointer boundaries for dynamic graceful restart capability
    - Add missing returns when parsing enhanced route-refresh
    - Return original as-path when reconciling AS versus AS4
    - Do not process route-refresh for AFI/SAFI if it's not negotiated
    - Check if we are not overusing error_data buffer when unknown cap received
    - fix NHT for explicit link-local BGP peers
    - improve packet parsing for EVPN and ENCAP/VNC
    - Prevent heap use-after-free for tunnel encapsulation attribute
    - Return 0 if AS4 capability is malformed
    - close dynamic peer socket in ttl error path
    - fix logic handling EVPN_FLAG_DEFAULT_GW
    - avoid early return in MPLSVPN NLRI processing
    - remove unneeded asserts in packet reads

eigrpd:
    - fix byte order in Hello TLV decode functions
    - Handling for malformed update packets
    - enforce minimum TLV length in Hello handler
    - reject invalid prefix mask len
    - skip unknown and ignored TLVs
    - Improve packet validation

isisd:
    - Reject SRv6 Locator TLV with Loc-Size of zero
    - consume leftover bytes after FAD sub-sub-TLV loop
    - use correct min size values for srv6 subtlvs
    - improve validation of flex-algo decoder
    - Fix missing neighbor address Sub-TLVs after link-params change
    - add unit test for remove_excess_adjs() memory leak fix
    - fix memory leak in remove_excess_adjs()
    - fix edge condition in max_lsp_count computation

ldpd:
    - improve tlv validation in several places

lib:
    - Report IPv6 MTU and not IPv4 for if_update_state_mtu6
    - disable warning in zlog.c to match master

nhrpd:
    - stop debugging auth credentials
    - fix byte-order when comparing error code in shortcut path
    - guard against zbuf_pulln NULL on truncated packets
    - require auth for all received packet types
    - harden debug packet parsing against malformed input
    - validate AFI index in extension replies

ospf6d:
    - move log call out of priv block
    - remove asserts in packet-handling paths
    - fix issues in ospf6 auth trailer code

ospfd:
    - add LSA validation in the apiserver path
    - add validation in several places before accessing

pceplib:
    - validate during of_list TLV decoding

pimd:
    - fix NOCACHE MFC resync detection log, add vrf name too
    - use upstream-owned pim pointer in register and upstream timers
    - use upstream-owned pim pointer in MSDP update paths
    - avoid JP build deref through channel OIL
    - guard RP RPF-failure mroute delete on detached OIL
    - avoid null deref in upstream delete debug path
    - guard channel OIL detach against stale pointers
    - fix crash due to double free
    - Ensure igmp message is of proper size
    - Reject pim packets with a malformed header length
    - Fix out of bounds read in AutoRP code
    - igmpv3 never checks packet length and trusts the num-sources field
    - Do not allow a register-stop message if not received from the RP
    - Prevent received msg length from being larger than buffer
    - Remove unnecessary asserts
    - When receiving a register stop ensure we have enough data to read
    - Ensure a register packet has enough space to read S,G data
    - Ensure that header has space on packet

ripngd:
    - fix data handling in several places

tests:
    - Check if route-map with set nexthop unchanged does not prevent outgoing
    - Check if mixed peer-group remote-as types can be used with auto
    - Verify neighbor addr Sub-TLVs after link-params reset
    - Expect return code being 0, not -1 when AS4 is empty or ASN is 0

vrrpd:

    - replace some asserts
    - only support ethernet in GARP code
    - limit advertised timers to 12-bits

## Checklist

- [x] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes


doc/ChangeLog.md
```diff
@@ -17,6 +17,7 @@ All notable changes to the project are documented in this file.
 - Handle unclean daemon exits better, e.g., `dbus-daemon` crashing and
   leaving a stale pidfile behind, causing it to refuse to be restarted
 - Fix occasional blank or garbled `[ OK ]` lines at startup
+- Fix MAC address input validation in infix-interfaces YANG.
```
troglobit: This is an important fix and needs a clear ChangeLog message.  "Fix" does not say what the change actually does, I suggest something along the lines of "Disallow multicast MAC addresses in custom MAC address configuration", or something to that effect.

troglobit (changes requested):

troglobit (approved):

PR #1509, opened 2026-05-19
Fix operational get-data crash on containers with shell chars

rical:
## Description

Yanger was populating a strict-pattern config leaf from podman's full entrypoint, breaking every operational read of containers whose command contained characters disallowed by the pattern. To fix this, we simply remove "command" from the data as "command" is config not operational.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [x] Test changes
- [ ] Build related changes

### DIFF
--- a/src/statd/python/yanger/infix_containers.py
+++ b/src/statd/python/yanger/infix_containers.py
@@ -234,10 +234,6 @@ def container(ps):
     if path:
         out["cmdline"] = " ".join([path] + args)
 
-    cmd = inspect.get("Config", {}).get("Cmd", [])
-    if cmd:
-        out["command"] = " ".join(cmd)
-
     net = network(ps, inspect)
     if net:
         out["network"] = net
--- a/test/case/statd/containers/infix-containers.json
+++ b/test/case/statd/containers/infix-containers.json
@@ -9,7 +9,6 @@
         "running": true,
         "status": "Up About a minute",
         "cmdline": "/usr/bin/tini -- /usr/sbin/httpd -f -v",
-        "command": "/usr/sbin/httpd -f -v",
         "network": {
           "interface": [
             {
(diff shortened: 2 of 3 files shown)


mattiaswal:
This feels like a bandaid fix, the core problem is that containers is not removed when change configuration to test config, right? Or what did you find out in your investigation @rical?

test/case/containers/cleanup/test.adoc
```diff
@@ -0,0 +1,26 @@
...
+==== Description
+
+Remove all containers and volumes from the target before running other
+container tests.  Persistent podman storage on test hardware can hold
+stale containers from earlier deployments or test runs, which then
+appear in operational data and may trip YANG validation -- making
+unrelated tests look flaky.
+
+This test runs first in the container suite to guarantee a known-empty
+podman state.
```
wkz: By introducing this test, are we not shielding ourselves from errors that our users may then face in a production setting instead?

Why do we sometimes have stale containers in `/var`? Isn't that the root cause?

wkz: Sorry, missed that @mattiaswal had already made this observation :smile:

rical:
> This feels like a bandaid fix, the core problem is that containers is not removed when change configuration to test config, right? Or what did you find out in your investigation @rical?

Yes, I totally agree. I rushed this. I'm normally the advocacy of aggressive pruning in the pursuit of immutability :) 

I will drop the commit from this PR. Rewrite the test to fail if it finds dangling containers and submit it in another PR, so we can merge the actual fix in this PR.

wkz (approved):
Love a fix that consists entirely of deletions 🤩

Issue #1301, opened 2025-12-03
Add support for unattended upgrades


mattiaswal:
### Description

Run a script in cron defined by user settings in the YANG model.

Example YANG:
```
  augment "/sys:system" {
    container unattended-upgrades {
      description
        "Configuration for automatic system software updates.

        When enabled, the system will automatically check for and
        install software upgrades at the specified time.";

      leaf enable {
        type boolean;
        default false;
        description
          "Enable or disable unattended upgrades.

          When set to true, the system will automatically check for
          available software updates and install them at the configured
          time.";
      }

      leaf time {
        type string {
          pattern '[0-2][0-9]:[0-5][0-9]';
        }
        mandatory true;
        description
          "Time of day to run the upgrade check and installation.

          Specified in 24-hour format as HH:MM (e.g., '02:00' for 2 AM,
          '14:30' for 2:30 PM).";
      }

      leaf-list day-of-week {
        type enumeration {
          enum "monday";
          enum "tuesday";
          enum "wednesday";
          enum "thursday";
          enum "friday";
          enum "saturday";
          enum "sunday";
        }
        description
          "Days of the week to run unattended upgrades.

          If not specified, upgrades will run daily. Multiple days
          can be selected to run upgrades only on specific days of
          the week.";
      }
    }
  }

```

### General Information

Anyone can help out by [sponsoring][1] development of new features or [contributing][2] pull requests.
Please use this issue for discussions related to the feature.

[1]: https://github.com/kernelkit/infix/blob/main/.github/SUPPORT.md
[2]: https://github.com/kernelkit/infix/blob/main/.github/CONTRIBUTING.md

troglobit:
> Good point, new suggestion, using ietf-schedule, limited on what cron supports.
>   augment /sys:system:
>     +--rw schedules
>        +--rw schedule* [name]
>           +--rw name           string
>           +--rw enabled?       boolean
>           +--rw description?   string
>           +--rw (action)
>           |  +--:(predefined)
>           |  |  +--rw predefined-action?   identityref

So the idea here is to set up a `predefined-action` right?  First one being `check-upgrades` perhaps, and then `unattended-upgrade` and `unattended-fixes` (for patch-releases only)?

wkz:
Wouldn't you have a single `predefined-action` ("`unattended-updates`"), which would mean "at the declared rate, check for and update software according to my policy in `/system/software/unattended-updates`?

[...]

PR #1523, opened 2026-06-02
Add support for ietf-schedule

saba8814:
### Why
Infix did not have scheduling system, thus this PR adds support for ietf-schedule(RFC9922) standard yang that is wired into the system using crontab. Additionally this PR adds 2 predefined actions that are check-update and reboot.

### What to Expect
Scheduler (infix-schedule) — a new YANG module augmenting ietf-system with a schedules list. Each schedule has a recurrence rule and a predefined action. A regression test testing scheduling of reboot predefined-action. Example configuration:

```json
{
  "ietf-system": {
    "system": {
      "infix-schedule:schedules": {
        "schedule": [
          {
            "name": "reboot-test",
            "enabled": true,
            "predefined-action": "infix-schedule:reboot",
            "recurrence": {
              "frequency": "infix-schedule:minutely",
              "interval": 1
            }
          }
        ]
      }
    }
  }
}
```

Firmware update check — when check-update fires, /usr/sbin/infix-check-update queries the GitHub releases API, compares the latest tag against the running version, and writes a notification to /run/infix-update. The login banner picks this up and displays it to the admin.

image

image

image

### Limitations
- Secondly and Yearly recurrence are not supported, since we can not express them in cron
- bymonthday only supports positive integer values 1-31 (ietf-schedule ranges are -31..-1 and 1-31)
- Two predefined-actions defined and implemented, custom command ommitted for now
- Check update script expects github api style

### Future work
- Add support for full-fledged unattended upgrades as discussed in #1301 

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [x] Feature
  - [x] Regression tests added?
  - [ ] Documentation added?
- [x] Test changes
- [ ] Build related changes


package/confd/crond.conf
```diff
@@ -0,0 +1,2 @@
+# Cron daemon for infix-schedule
+service [S12345] crond -f -- Cron daemon
```
mattiaswal: Do not start in runlevel S, 2345 should be enough. This since we have seen slow platforms (arm32) struggle with starting to much, when it should focus on confd.

src/confd/src/schedule.c
```diff
@@ -0,0 +1,225 @@
...
+	freq  = lydx_get_cattr(recurrence, "frequency");
+	ivstr = lydx_get_cattr(recurrence, "interval");
+	if (!freq || !ivstr)
+		goto done;
+
+	iv = atoi(ivstr);
+	if (iv <= 0)
+		iv = 1;
+
+	if (strstr(freq, "secondly")) {
```
mattiaswal: If it not possible to set in cron, it should not be possible to set in YANG, block it in yang model, the problematic here is that it is an identity, do not remember how we did with unsupported identities. Lets discuss this.

We have as a policy that everything should be validated in the yang model (using deviate or must expression, mandatory etc), to keep the c-code as little as possible.

src/confd/yang/confd/infix-schedule@2026-05-27.yang
```diff
@@ -0,0 +1,108 @@
...
+    base predefined-action;
+    description
+      "Check for and notify about available firmware updates.
+
+       The update policy is configured under
+       /ietf-system:system/infix-system:software/check-update.";
+  }
+
+  augment "/sys:system" {
+    container schedules {
```
mattiaswal: Missing "description"

src/confd/yang/confd/infix-schedule@2026-05-27.yang
```diff
@@ -0,0 +1,108 @@
...
+          type string;
+        }
+        leaf enabled {
+          type boolean;
+          default "true";
+        }
+        leaf description {
+          type string;
+        }
+        choice action {
```
mattiaswal: Missing description

src/confd/yang/confd/infix-system-software.yang
```diff
@@ -112,6 +148,30 @@ submodule infix-system-software {
...
+
+        leaf url {
+          type string;
+          description
+            "Direct download URL for the available bundle.";
+        }
+
+        leaf last-check {
+          type yang:date-and-time;
+          description
```
mattiaswal: For things that pure operational, set as config: false

src/confd/yang/confd/infix-system-software@2026-05-29.yang
```diff
@@ -0,0 +1,290 @@
+submodule infix-system-software {
```
mattiaswal: This file should just be a symlink to infix-system-software.yang, see the surrounding files

src/confd/yang/confd.inc
```diff
@@ -57,4 +57,6 @@ MODULES=(
 	"ieee1588-ptp-tt@2023-08-14.yang -e timestamp-correction"
 	"ieee802-dot1as-gptp@2025-12-10.yang"
 	"infix-ptp@2026-04-07.yang"
+	"ietf-schedule@2026-03-10.yang -e basic-recurrence"
+	"infix-schedule@2026-05-27.yang"
 )
```
mattiaswal, line 62: In this file you also need to update the reference to all updated yang models, in this case infix-system-software@.....yang

test/case/system/schedule_reboot/test.py
```diff
@@ -0,0 +1,47 @@
...
+                                    "interval": 1
+                                }
+                            }
+                        ]
+                    }
+                }
+            }
+        })
+
+    with test.step("Verify schedule is configured"):
```
mattiaswal: This is redundat, since if the YANG model succeed to apply, it is configured.

src/confd/yang/confd/infix-schedule@2026-05-27.yang
```diff
@@ -0,0 +1,108 @@
+module infix-schedule {
```
mattiaswal: this file should just be a symlink to infix-schedule.yang, see surrounding files.

board/common/rootfs/usr/sbin/infix-check-update
```diff
@@ -0,0 +1,58 @@
...
+. /etc/os-release
+
+# Dev/dirty builds have no comparable semver — always show the latest release
+IS_RELEASE=true
+if ! echo "$VERSION" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+'; then
+    IS_RELEASE=false
+fi
+
+# Read configured update-url from sysrepo, fall back to upstream
+UPDATE_URL=$(sysrepocfg -d running -f json \
```
mattiaswal: Use jq here instead of calls to misc shell programs, this to limit forking:
like this (This is pure AI, not tested, just to give you a hint what i mean):
                                                                                                                 
     UPDATE_URL=$(sysrepocfg -d running -f json \                                                                                           
         -x '/ietf-system:system/infix-system:software/check-update/update-url' \                                                                            
         2>/dev/null \                                                                                                                                       
         | jq -r '.. | objects | ."update-url"? // empty')

mattiaswal: I think you should look to use the 'copy' shell command instead of sysrepocfg here, since there will be problem calling sysrepocfg for everone except root:
```
admin@infix-00-00-00:~$ sysrepocfg -X -f json -d operational
[ERR] open() on "/etc/sysrepo/sr_main_lock" failed (Permission denied).
sysrepocfg error: Failed to connect (System function call failed)
admin@infix-00-00-00:~$ 
```
copy also respect nacm if i remember it correctly.

mattiaswal (changes requested):

mattiaswal:
Also add the the "magic" key word fix #issue number in the PR description, this closes the open issue automatically when merged.

src/confd/src/schedule.c
```diff
@@ -0,0 +1,225 @@
...
+ *   monthly/N  → 0 0 1 *\/N *
+ *
+ * Optional by-* leaves refine the expression:
+ *   byminute → replaces the minute field
+ *   byhour   → replaces the hour field
+ *   byday    → replaces the day-of-week field
+ */
+static void build_cron_expr(struct lyd_node *recurrence, char *expr, size_t sz)
+{
+	const char *freq, *ivstr;
```
mattiaswal, line 58: coding style: We use reverse Christmas tree for variable definition.

saba8814:
> Also add the the "magic" key word fix #issue number in the PR description, this closes the open issue automatically when merged.

Thing is #1301  is formulated in way that it is support for full unattended upgrade, and this PR adds scheduling stuff and just a check for update availiability thus I ommited it from git message, since full implementation of unnattended-upgrade will be handled in a separete merge request.

troglobit:
Still haven't had the time to fully look at this, but a few general pointers that are useful:

 1. Personally I prefer to use "Add support for" instead of "feat:", it feels more human and invites to a discussion
 2. Each PR *should* have a description which introduces the reviewer to the topic, nothing too in-depth, just a summary of the why:s and what:s to expect.  Screenshots is a favorite!  This is also a good place to bring up limitations (for example: "most of the advanced features are deviated to allow for a simple good-enough first release", future work, etc.

I hope to have some time tomorrow for a detailed look at the changes :smiley:

saba8814:
> Still haven't had the time to fully look at this, but a few general pointers that are useful:
> 
> 1. Personally I prefer to use "Add support for" instead of "feat:", it feels more human and invites to a discussion
> 2. Each PR _should_ have a description which introduces the reviewer to the topic, nothing too in-depth, just a summary of the why:s and what:s to expect.  Screenshots is a favorite!  This is also a good place to bring up limitations (for example: "most of the advanced features are deviated to allow for a simple good-enough first release", future work, etc.
> 
> I hope to have some time tomorrow for a detailed look at the changes 😃

Thank you, I've added description now, hope it is not too long.

board/common/rootfs/usr/sbin/infix-check-update
```diff
@@ -0,0 +1,56 @@
...
+    | jq -r '.. | objects | ."update-url"? // empty')
+UPDATE_URL=${UPDATE_URL:-"https://github.com/kernelkit/infix"}
+
+# Derive API URL from the configured update URL.
+# Default (github.com): https://github.com/org/repo → https://api.github.com/repos/org/repo
+REPO=$(echo "$UPDATE_URL" | sed 's|https://github.com/||; s|/*$||')
+API_URL="https://api.github.com/repos/${REPO}/releases/latest"
+
+LATEST_TAG=$(curl -sSL --max-time 10 "$API_URL" 2>/dev/null \
+    | sed -n 's/.*"tag_name" *: *"\([^"]*\)".*/\1/p' \
```
mattiaswal: Use jq instead:
lazzer@tollan ~/src/github.com/kernelkit/infix4 (wifi-roaming)$ curl -sSL --max-time 10  https://api.github.com/repos/kernelkit/infix/releases/latest | jq -r '.tag_name'
v26.05.0
lazzer@tollan ~/src/github.com/kernelkit/infix4 (wifi-roaming)$

src/confd/src/schedule.c
```diff
@@ -0,0 +1,242 @@
...
+		const char *name, *action, *cmd;
+		char expr[128];
+
+		if (!lydx_is_enabled(sched, "enabled"))
+			continue;
+
+		name = lydx_get_cattr(sched, "name");
+
+		recurrence = lydx_get_child(sched, "recurrence");
+		if (!recurrence)
```
mattiaswal: Mark the container `recurrence` mandatory instead of checking if it is set in the c-code

src/confd/src/schedule.c
```diff
@@ -0,0 +1,242 @@
...
+			continue;
+
+		build_cron_expr(recurrence, expr, sizeof(expr));
+
+		action = lydx_get_cattr(sched, "predefined-action");
+		cmd = action_to_cmd(action);
+		if (!cmd)
```
mattiaswal: We should not get here, log if we do with ERROR()

src/confd/src/schedule.c
```diff
@@ -0,0 +1,242 @@
...
+int schedule_init(struct confd *confd)
+{
+	sr_data_t *data = NULL;
+
+	/*
+	 * Sync the crontab with current running config at startup so
+	 * scheduled jobs survive across reboots.
+	 */
+	sr_get_data(confd->session, "//.", 0, 0, 0, &data);
+	apply_schedules(data ? data->tree : NULL);
```
mattiaswal: Subscribe on SR_EV_ENABLED instead of running code in the init call.

src/confd/yang/confd/infix-schedule.yang
```diff
@@ -0,0 +1,217 @@
...
+          }
+          leaf-list byyearmonth {
+            type uint32 {
+              range "1..12";
+            }
+            description
+              "Restrict firing to specific months of the year (1-12).";
+          }
+        }
+        container status {
```
mattiaswal: Lacking the yanger.py code to implement this node.

src/confd/yang/confd/infix-system-software.yang
```diff
@@ -80,6 +84,38 @@ submodule infix-system-software {
         "The last error encountered by the installer service.";
     }
   }
+  augment "/sys:system" {
+    container software {
+      description
+        "Software management configuration.";
+
+      container check-update {
```
mattiaswal, line 96: Shouldn't this setting be inside the schudler instead? What do you say @wkz  @troglobit

mattiaswal (changes requested):

src/confd/src/schedule.c
mattiaswal: I had a second thougth on this, this should not be needed at all to run here, right? It gets run when the changes happen

board/common/rootfs/usr/sbin/infix-check-update
troglobit: Very limited review, I think we should also just display the "dirname" of the release. Makes for a shorter and more succinct notification in the cli and WebUI.

src/confd/yang/confd/infix-system-software.yang
troglobit: Quite possibly, please remind me again of this tomorrow. 

Sorry for being so late on the ball!

saba8814: I've cleaned up other comments, I will wait on this one until I hear your point of view.  I believe it shall be under software yang since we might have "action" in future to manually check for update. FYI: @troglobit  @mattiaswal

troglobit: I think this is OK, but I have another point to make that will tie actions like this back to the scheduler better.

troglobit (changes requested):
This is shaping up!  I have two architectural points to bring up: the intended pattern in the RFC is to pick the grouping that matches our capability and _prune_, then bring the schedule *to* the task rather than the task to a central scheduler.

We have sort of done this, but I think we can do better! 😃 

### 1. Use `icalendar-recurrence`, prune with deviations

Right now we `uses schedule:recurrence-basic` and hand-declare `byhour`, `byminute`, `byday`, etc. on top. But those nodes already exist in the RFC's `icalendar-recurrence` grouping, with semantics pinned to RFC 5545. Declaring them locally means a generic RFC 9922 client sees `infix-schedule:byhour` instead of `ietf-chedule:byhour`, we pay for the standard without getting the interop. So, take the top grouping and deviate away what cron can't do:

```yang
container recurrence {
  if-feature "schedule:icalendar-recurrence";
  uses schedule:icalendar-recurrence {
    refine frequency {
      mandatory true;   // RFC 9922 §3.3.3 SHOULD; also stops an empty
                        // recurrence from compiling to '* * * * *' (!)
      must "not(derived-from-or-self(., 'schedule:secondly'))";
    }
    refine interval { default 1; }
  }
}

deviation ".../recurrence/recurrence-first"     { deviate not-supported; }
deviation ".../recurrence/until"                { deviate not-supported; }
deviation ".../recurrence/count"                { deviate not-supported; }
deviation ".../recurrence/period"               { deviate not-supported; }
deviation ".../recurrence/bysecond"             { deviate not-supported; }
deviation ".../recurrence/byday/direction"      { deviate not-supported; }
// ... byyearday, byyearweek, bysetpos, workweek-start, exception-dates,
//     time-zone-identifier — same treatment
deviation ".../recurrence/bymonthday" {
  deviate replace { type int32 { range "1..31"; } }
}
```

What survives is exactly the surface you built — same feature set, standard spelling. Bonuses: the `supported-frequency` identities and both existing deviations go away (clients configure plain `ietf-schedule:daily`, the `must` rejects what we can't do), deviations are advertised in the YANG library so controllers can discover our limits, and when we someday outgrow crond we *lift* deviations instead of renaming nodes.

### 2. Drop `choice action` — schedules become named time-specs, actions reference them

The `predefined-action` identityref bothers me: the schedule holds the trigger, but check-update's actual config lives in `/system/software/ check-update`, and nothing in the schema connects them — just a naming convention and a `strstr()` in `schedule.c`. Two `enabled` leaves with undefined interaction, and every new action needs an identity here plus a branch in `action_to_cmd()`. The scheduler ends up knowing about everything.

This is the old Cisco `time-range` pattern: make the schedules list a library of pure *time definitions* (no actions at all), and let each feature own its trigger via a YANG leafref:

```yang
/* infix-schedule: just time-specs */
typedef schedule-ref {
  type leafref {
    path "/sys:system/schedules/schedule/name";
  }
}

/* infix-system-software: the feature owns its complete self */
container check-update {
  leaf enabled    { type boolean; default false; }
  leaf update-url { type string; ... }
  leaf schedule   { type ifsched:schedule-ref; }
}

/* scheduled reboot becomes its own small feature */
container scheduled-reboot {
  leaf schedule { type ifsched:schedule-ref; }
}
```

This gets us validated references (can't delete a schedule still in use, nice when the action is *reboot*), clean `enabled` semantics (feature's `enabled` = whether, schedule's `enabled` = kill-switch for everything referencing it), sharing (`reboot` and `check-update` both pointing at `sunday-night` is one recurrence to audit), and new schedulable features never touch infix-schedule again. `schedule.c` shrinks to "schedule name → cron fields"; each feature's callback knows its own command, so `action_to_cmd()` and the identities go away too. This also sets us up nicely for #1301, where unattended upgrades will grow real policy config around this trigger.

One backend caveat to handle either way: with both day-of-month and
day-of-week set, cron fires on *either*, while RFC 5545 says intersection, so `bymonthday 13` + `byday friday` means Friday the 13th in the model but "every 13th and every Friday" in the crontab. Simplest fix: `must "not(bymonthday and byday)"` until someone needs the combination.

I know this is a rework rather than a touch-up, but the cron translation layer you've built carries over almost unchanged, it's only the YANG surface that changes, but this is very important and the leafref addition means it will be possible to validate the config against the YANG model offline!

board/common/rootfs/usr/sbin/infix-check-update
```diff
@@ -0,0 +1,55 @@
+#!/bin/sh
+# Check for available software updates and notify on login if one exists.
+# Called by the scheduler.
+
+NOTIFY_FILE=/run/infix-update
+TAG=infix-update
```
troglobit: Suggest we rename these to os-update.  Avoids any future branding issues.

This is a general recommendation, even though the YANG models have 'infix' in their name, we try to abstain for encoding it in commands and state files.

board/common/rootfs/usr/sbin/check-update
troglobit, line 1: Please rename this to check-update, like the command itself.

board/common/rootfs/etc/profile.d/update-check.sh
```diff
@@ -0,0 +1,3 @@
+if [ -s /run/infix-update ]; then
+    printf '\n\033[1;33m *** %s ***\033[0m\n\n' "$(cat /run/infix-update)"
```
troglobit: Rename to /run/os-update

board/common/rootfs/etc/tmpfiles.d/infix-schedule.conf
troglobit, line 1: Rename to os-schedule.conf and content to /run/os-update

src/confd/src/system-software.c
```diff
@@ -89,6 +89,14 @@ static int infix_system_sw_set_boot_order(sr_session_ctx_t *session, uint32_t su
 	return SR_ERR_OK;
 }
 
+/* Scheduler consumer for check-update. */
+static const struct cron_consumer check_update_consumer = {
+	.path	      = "/ietf-system:system/infix-system:software/check-update",
+	.sched_leaf   = "schedule",
+	.enabled_leaf = "enabled",
+	.command      = "/usr/sbin/infix-check-update",
```
troglobit: This the becomes `/usr/sbin/check-update`

src/webui/internal/handlers/dashboard.go
```diff
@@ -305,6 +311,31 @@ type diskEntry struct {
 // reachability row — a well-known, stable anycast resolver.
 const internetProbe = "1.1.1.1"
 
+// updateNoticeFile holds the software-update notice written by
+// /usr/sbin/infix-check-update — the same text the CLI shows at login.
+// A package var so tests can point it elsewhere.
+var updateNoticeFile = "/run/infix-update"
```
troglobit: Rename to /run/os-update

troglobit (changes requested):
A few more minor changes, but now we're really close. The YANG model looks great!

troglobit (approved):
Perfect, thanks for hanging in there!

troglobit:
@mattiaswal I think this PR is only blocked by your outstanding review since you requested changes a while back. (It wasn't enough just for me to approve ...)

mattiaswal (approved):

PR #1539, opened 2026-06-16
Web test check

troglobit:
## Description

Minor fixes, and a diff view!

## Checklist

- [x] Bugfix
  - [ ] Regression tests
- [x] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes


test/case/services/webui/login_csrf/test.py
```diff
@@ -102,6 +102,13 @@ def login(self):
         host = target.location.host
         password = target.location.password
 
+        # The web UI is an optional package (BR2_PACKAGE_WEBUI); minimal
+        # builds omit it.  Skip rather than fail when it isn't installed.
+        tgtssh = env.attach("target", "mgmt", "ssh")
```
mattiaswal: Hmm, shouldn't this be a feature on the infix-services yang model and feature in infix? I think this is the way we want to go.

mattiaswal (changes requested):

test/case/services/webui/login_csrf/test.py
mattiaswal: If it is even possible, previously we have on checked if a module exist.

mattiaswal: Looks like it is a has_feature in transport.py :+1:

troglobit: Tricky question, previously we've had a simple landing page shown at :80/443 when the web interface was enabled, for both minimal and regular builds. Now that landing page is only shown for minimal builds.  So, one approach could be to add a conditional /web/management/ container that is only included when the webui is built.  Another could be a take on my earlier proposal to make the minimal build *really* minimal and make the entire /web section optional, i.e., no nginx or restconf even.

troglobit: We could accept this change as-is now, so we fix the test issue of all branches from main, and open a discussion on the services feature as a separate topic?

mattiaswal: Just skipping the UI (management node as you suggest) is enough. Dropping restconf from minimal is still a bad idea I think since we use it for tests. Dropping restconf I feel will be to much loss of test coverage.

mattiaswal: Another approach is that you just can have a feature enabled without limit any containers in the model

troglobit: Not sure I follow, what does "have a feature enabled without limit any containers in the model"?

Currently we have a LOT of enabled flags in the `/web` service tree:
```
module: infix-services
  +--rw web
     +--rw certificate?   ks:central-asymmetric-key-ref
     +--rw enabled?       boolean
     +--rw console
     |  +--rw enabled?   boolean
     +--rw netbrowse
     |  +--rw enabled?   boolean
     +--rw restconf
        +--rw enabled?   boolean
```
Personally I feel that `/web/enabled` is the logical knob for users to toggle to enable/disable the webui, adding a `/web/management/enabled` knob will be confusing for sure.

troglobit: Wait, @mattiaswal, is it a *YANG* feature you're talking about?

mattiaswal (approved):
Great work! Looking forward to do some additions myself.

troglobit:
Only one unrelated test case on one system failing (WireGuard Multipoint). Merging.

PR #1541, opened 2026-06-17
Use kernelkit/goyang instead of patched builtin

troglobit:
## Description

Refactor, or cleanup, of one of the new webui's dependencies.  This drops the locally patched "fork" of goyang with the same open pattern used by KernelKit for other deps.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes


.github/workflows/govulncheck.yml
```diff
@@ -0,0 +1,65 @@
...
+          # whenever it finds anything, so don't let it fail the step here.
+          {
+            echo "## govulncheck: ${{ matrix.module }}"
+            echo '```'
+            govulncheck ./... || true
+            echo '```'
+          } | tee -a "$GITHUB_STEP_SUMMARY"
+
+          # Gate only on vulnerabilities our code actually calls in a
+          # dependency.  Called stdlib vulns are fixed by bumping the
```
mattiaswal: Does this handle indirect usage as well? If we use a function that uses something bad?

troglobit: The comment is a bit vague, `govulncheck` actually reaches any vulnerable symbol in the call graph. I've updated the comment.

mattiaswal (approved):

PR #1545, opened 2026-06-23
test: add generic boot-from-factory-config test

rical:
## Description

Verify that the device’s factory-default configuration boots cleanly and that the device remains usable afterwards. I.e. it does not fall back to the fail-secure failure-config.

This exercises the device’s own first-boot bootstrap path: with no startup-config present, confd initialises running from the factory-config. That is exactly what a factory-fresh (or factory-reset) device does, and it avoids applying a full config swap over the live management session.

The test is image-generic: it uses whatever factory-config the running image was built with, so it covers both the stock Infix factory config and any spin factory config.

A single-node topology is used on purpose: a factory config is not written with a lab full of peers in mind, so applying it across a multi-node topology could trigger broadcast storms or similar.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [x] Test changes
- [ ] Build related changes


troglobit:
Assigning to @wkz for review. Would be great to have this merged for the upcoming release.

test/case/system/factory_config/test.py
```diff
@@ -0,0 +1,67 @@
...
+def factory_hostname(tgtssh):
+    """Read the hostname the factory-config will boot with."""
+    cfg = json.loads(tgtssh.runsh(f"cat {FACTORY}").stdout)
+    return cfg.get("ietf-system:system", {}).get("hostname")
+
+
+def cleanup(env):
+    """Restore the rig to the clean per-test baseline for the next test."""
+    print("Restoring device to clean test baseline")
+    target = env.attach("target", "mgmt", "netconf")
```
wkz:
```suggestion
    target = env.attach("target", "mgmt")
```
I believe this will allow the test to be run both with NETCONF _and_ RESTCONF, according to the regular randomization logic.

test/case/system/factory_config/test.py
```diff
@@ -0,0 +1,67 @@
...
+    target.reboot()
+    if not wait_boot(target, env):
+        test.fail("Device did not come back while restoring baseline")
+
+
+with infamy.Test() as test:
+    with test.step("Set up topology and attach to target DUT"):
+        env = infamy.Env()
+        target = env.attach("target", "mgmt", "netconf")
```
wkz:
```suggestion
        target = env.attach("target", "mgmt")
```
Same here.

test/case/system/factory_config/test.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def factory_hostname(tgtssh):
+    """Read the hostname the factory-config will boot with."""
+    cfg = json.loads(tgtssh.runsh(f"cat {FACTORY}").stdout)
+    return cfg.get("ietf-system:system", {}).get("hostname")
+
+
+def cleanup(env):
+    """Restore the rig to the clean per-test baseline for the next test."""
```
wkz: What exactly are we restoring with the extra reboot?

test/case/system/factory_config/test.py
```diff
@@ -0,0 +1,67 @@
...
+        target.startup_override()
+
+    with test.step("Reboot onto the factory config"):
+        target.reboot()
+        if not wait_boot(target, env):
+            test.fail("Device did not boot from factory config")
+        test.push_test_cleanup(lambda: cleanup(env))
+
+    with test.step("Verify device is usable and not in failure-config"):
+        target = env.attach("target", "mgmt", "netconf", test_reset=False)
```
wkz:
```suggestion
        target = env.attach("target", "mgmt", test_reset=False)
```

wkz (changes requested):
Very nice to not have to worry about these kinds of failures in the future! 😅

test/case/system/factory_config/test.py
rical: We are reverting from factory-config back to test-config

wkz (approved):

PR #1573, opened 2026-08-02
Revert statd fix that cause massive test regressions

troglobit:
## Description

The statd mDNS neighbor collector sometimes locks up and times out with:

Aug  2 07:11:53 test-00-01-00 statd[3558]: mdns: sr_apply_changes: Timeout expired
Aug  2 07:11:53 test-00-01-00 statd[3558]: Error, getting operational data: User callback failed

With the extended sysrepo API timeout introduced in c70a699, PR #1572, this causes massive test regressions, hence this revert.

## Checklist

Tick *relevant* boxes, this PR is-a or has-a:

- [ ] Bugfix
  - [ ] Regression tests
  - [ ] ChangeLog updates (for next release)
- [ ] Feature
  - [ ] YANG model change => revision updated?
  - [ ] Regression tests added?
  - [ ] ChangeLog updates (for next release)
  - [ ] Documentation added?
- [ ] Test changes
  - [ ] Checked in changed Readme.adoc (make test-spec)
  - [ ] Added new test to group Readme.adoc and yaml file
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (please detail in commit messages)
- [ ] Build related changes
- [ ] Documentation content changes
  - [ ] ChangeLog updated (for major changes)
- [x] Other (please describe): revert for massive regressions

PR #1574, opened 2026-08-02
Stability fixes

troglobit:
## Description

Follow-up to the `statd` timeout revert (#1573), fixing the actual root causes behind the regression failures on main, plus related cleanups found along the way.

The meltdown had four layers: `statd` blocking its event loop (the reverted trigger), the journal snapshot monopolizing the datastore for minutes, mDNS browse storms feeding more contention, *and* interrupted commits leaving dagger generations half-applied in the kernel, permanently breaking all subsequent interface configuration until reboot.

- `statd`: mDNS neighbor updates are debounced into one apply per burst and retried with backoff on datastore contention, the event loop never blocks on a busy datastore
- `statd`: journal snapshots run in a forked, low-priority statd-journal process with its own sysrepo connection, reading one module per call so other datastore users interleave.  The timer is re-armed only after completion, so snapshots never overlap.  Drops the last thread, and the second sysrepo session, from `statd`
- `confd`: interface setup/teardown is now convergent, leftovers from a partially applied generation are removed before link add, and deleting an already-missing interface no longer aborts the generation
- `confd`: enabling an optional service not present in the image no longer creates a dangling Finit symlink; build also drops leftover confs (ttyd) and prunes dangling symlinks, silencing minimal builds
- `statd`: proper option handling (`-h`, `-V`, `-v <level>`) backed by `setlogmask()`, defaulting to `NOTICE` like other services; matching `-v info/debug` fixes in `confd`; `yanger` errors now name the module/interface and exit code
- [mdns-alias v1.3](https://github.com/troglobit/mdns-alias/releases/tag/v1.3): fixes crash when a hostname change coincides with an avahi restart ("Failed creating new entry group: Bad state")

## Checklist

- [x] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes


doc/ChangeLog.md
```diff
@@ -10,15 +10,51 @@ All notable changes to the project are documented in this file.
...
 
 - Fix annoying "cannot deselect all services" or reset to YANG default in the
   web interface's firewall configuration page
+- Fix sporadic slow response, or timeouts, when reading device status while
+  mDNS neighbors are being discovered, e.g., after an mDNS restart.  Updates
+  to the neighbor table are now batched, and politely retried when other
+  users or services keep the system busy, logged as:
+
+        statd[3558]: mdns: sr_apply_changes: Timeout expired
+
```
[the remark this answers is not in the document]
troglobit: Fixed locally, pushing in a minute.

package/statd/statd.conf
```diff
@@ -1,2 +1,2 @@
 #set DEBUG=1
-service name:statd [12345] <pid/confd> statd -f -p /run/statd.pid -n -- Status daemon
+service name:statd [12345] <pid/confd> statd -- Status daemon
```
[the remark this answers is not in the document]
troglobit: Fixed locally, pushing in a minute.

src/statd/journal.c
```diff
@@ -1,31 +1,45 @@
...
+#include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <unistd.h>
-#include <sysrepo.h>
-#include <ev.h>
 #include <string.h>
-#include <errno.h>
-#include <time.h>
+#include <sys/prctl.h>
```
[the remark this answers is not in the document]
troglobit: Fixed locally, pushing in a minute.

board/common/post-build.sh
```diff
@@ -146,3 +146,14 @@ mkuserguide()
 if [ "$BR2_PACKAGE_WEBUI" = "y" ]; then
     mkuserguide
 fi
+
```
mattiaswal: Isn't a better solution to move the installation of the ttyd.conf to the ttyd package instead?

src/confd/src/interfaces.c
```diff
@@ -507,13 +507,26 @@ static int eth_gen_del(struct lyd_node *dif, FILE *ip)
 	return 0;
 }
 
-static int link_gen_del(struct lyd_node *dif, FILE *ip)
+/*
```
mattiaswal: This we have talked multiple times about, i am fine with this solution, but i remember @wkz had some good examples why this was not good.

src/confd/src/interfaces.c
```diff
@@ -617,6 +627,49 @@ static sr_error_t netdag_gen_iface_timeout(struct dagger *net, const char *ifnam
 	return SR_ERR_OK;
 }
 
+/*
```
mattiaswal: We need to discuss this next week. What should happen when something fails, should we handle it or should it never fail?

src/confd/src/interfaces.c
wkz: Not exactly sure what example I was thinking about.

My main concerns about this is (1) that we mask the sources of errors (i.e., if there is a problem in removing an interface, we won't see it until we try to configure the next instance of it - or worse: never); and (2) that we lull the user into thinking we support things (rolling back a partially applied config) that we do not (i.e., we might still be leaving addresses, routes, containers, etc. around from the previous config). Unless we have that sorted out through some other means?

src/confd/src/interfaces.c
troglobit: Team decision to drop this commit.  Will push an updated branch soon.

src/confd/src/interfaces.c
troglobit: Team decision to drop this commit. Will push an updated branch soon.

board/common/post-build.sh
troglobit: Well, `ttyd` is packaged in Buildroot so that won't fly, but I could relocate this fixup to our `package/skeleton-init-finit/` instead.  That's a cleaner fit and we reduce the clutter in the post-build script.

board/common/post-build.sh
```diff
@@ -146,3 +146,8 @@ mkuserguide()
...
+# Drop dangling Finit enabled/*.conf symlinks, e.g., optional services
+# not part of this image, they cause noise at every initctl reload.
+# NOTE: must be the last step before creating the image!
+find "$TARGET_DIR/etc/finit.d/enabled" -xtype l -delete 2>/dev/null
```
mattiaswal: This should be fixed once, not on each build. Increase the build-time unnecessary.

mattiaswal (changes requested):

board/common/post-build.sh
troglobit: Reworked, hope this sits better with you.

mattiaswal (approved):

PR #1586, opened 2026-08-14
confd/board: nanopi-r2s: add gencert keystore entry to factory config

pjator:
Since the HTTPS certificate moved into the ietf-keystore, the Web UI (nginx) is broken on fresh installs: no factory config ships the "gencert" entry, so confd never generates the self-signed certificate, /etc/ssl/certs/self-signed.crt is missing and nginx refuses to start. Add the entry (empty keys, generated on first boot) and reference it from infix-services:web.certificate.

## Checklist

Tick *relevant* boxes, this PR is-a or has-a:

- [ x] Bugfix
  - [ ] Regression tests
  - [ ] ChangeLog updates (for next release)
- [ ] Feature
  - [ ] YANG model change => revision updated?
  - [ ] Regression tests added?
  - [ ] ChangeLog updates (for next release)
  - [ ] Documentation added?
- [ ] Test changes
  - [ ] Checked in changed Readme.adoc (make test-spec)
  - [ ] Added new test to group Readme.adoc and yaml file
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (please detail in commit messages)
- [ ] Build related changes
- [ ] Documentation content changes
  - [ ] ChangeLog updated (for major changes)
- [ ] Other (please describe):

PR #1587, opened 2026-08-14
confd: set IITO startup/failure conditions (fixes status LEDs)

pjator:
iitod.json expects /run/finit/cond/run/startup/success before the LAN and status LEDs light up, but nothing ever creates it.  Add a small finit service (after confd/ready) that sets the condition on every boot.

## Checklist

- [x] Bugfix
  - [ ] Regression tests
- [ ] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes

### DIFF
--- a/src/confd/src/main.c
+++ b/src/confd/src/main.c
@@ -49,6 +49,23 @@
  */
 #define SENTINEL_PATH "/run/confd.boot"
 
+/*
+ * Set a finit condition in the usr/ namespace, e.g.
+ * "usr/startup-config-ok", used to signal IITO (and finit services) about
+ * the outcome of the bootstrap/startup sequence.  Finit guarantees
+ * /run/finit/cond/usr exists (via its tmpfiles snippet), so no parent
+ * directories are created here.
+ */
+static void set_finit_cond(const char *cond)
+{
+	char path[128];
+	snprintf(path, sizeof(path), "/run/finit/cond/usr/%s", cond);
+
+
+	if (symlink("/run/finit/cond/reconf", path) && errno != EEXIST)
+		WARN("Failed to set finit condition %s: %m", path);
+}
+
 /* Callback type names from sysrepo plugin API */
 #define SRP_INIT_CB    "sr_plugin_init_cb"
 #define SRP_CLEANUP_CB "sr_plugin_cleanup_cb"
@@ -505,27 +522,32 @@ static void handle_startup_failure(sr_session_ctx_t *sess, const char *failure_p
 	int r;
 
 	ERROR("Failed loading startup-config, reverting to Fail Secure mode!");
+	set_finit_cond("startup-config-error");
 
 	/* Reset to factory-default */
 	r = sr_copy_config(sess, NULL, SR_DS_FACTORY_DEFAULT, timeout_ms);
 	if (r != SR_ERR_OK) {
 		ERROR("sr_copy_config(factory-default) failed: %s", sr_strerror(r));
 		/* Nuclear option: wipe everything */
+		banner_append("CRITICAL ERROR: Logins are disabled, no credentials available");
 		systemf("rm -f /etc/sysrepo/data/*startup* /etc/sysrepo/data/*running* /dev/shm/sr_*");
+		set_finit_cond("failure-config-error");
 		return;
 	}
 
 	/* Load failure-config on top */
 	if (fexist(failure_path)) {
 		if (load_config(conn, sess, failure_path, timeout_ms)) {
-			ERROR("Failed loading failure-config, aborting!");
+			ERROR("failed applying failure-config, aborting!");
 			banner_append("CRITICAL ERROR: Logins are disabled, no credentials available");
 			systemf("initctl -nbq runlevel 9");
+			set_finit_cond("failure-config-error");
 			return;
 		}
 	}
 
 	banner_append("ERROR: Corrupt startup-config, system has reverted to default login credentials");
+	set_finit_cond("failure-config-ok");
 }
 
 /*
@@ -587,6 +609,7 @@ static int bootstrap_config(sr_conn_ctx_t *conn, sr_session_ctx_t *sess,
 		if (r != SR_ERR_OK)
 			WARN("Failed to sync startup datastore: %s", sr_strerror(r));
 
+		set_finit_cond("startup-config-ok");
 		return 0;
 	}
 
@@ -841,7 +864,7 @@ int main(int argc, char **argv)
 	quiet_now();
 
 	/* Signal that bootstrap is complete (dbus, resolvconf depend on this) */
-	symlink("/run/finit/cond/reconf", "/run/finit/cond/usr/bootstrap");
+	set_finit_cond("bootstrap");
 
 	/*
 	 * Write restart sentinel.  From this point on, sysrepo and the system
--- a/board/aarch64/friendlyarm-nanopi-r2s/rootfs/usr/share/product/friendlyarm,nanopi-r2s/etc/iitod.json
+++ b/board/aarch64/friendlyarm-nanopi-r2s/rootfs/usr/share/product/friendlyarm,nanopi-r2s/etc/iitod.json
@@ -15,9 +15,9 @@
 
 	    "wan-up":       { "path": "/run/led/wan-up" },
 
-	    "startup":   { "path": "/run/finit/cond/run/startup/success" },
-	    "fail-safe": { "path": "/run/finit/cond/run/failure/success" },
-	    "panic":     { "path": "/run/finit/cond/run/failure/failure" }
+	    "startup":   { "path": "/run/finit/cond/usr/startup-config-ok" },
+	    "fail-safe": { "path": "/run/finit/cond/usr/failure-config-ok" },
+	    "panic":     { "path": "/run/finit/cond/usr/failure-config-error" }
 	},
 	"udev": {
 	    "power-a": { "subsystem": "power_supply" },
--- a/board/common/rootfs/etc/iitod.json
+++ b/board/common/rootfs/etc/iitod.json
@@ -18,9 +18,9 @@
 	    "lan-err":      { "path": "/run/led/lan-err" },
 	    "lan-crit":     { "path": "/run/led/lan-crit" },
 
-	    "startup":   { "path": "/run/finit/cond/run/startup/success" },
-	    "fail-safe": { "path": "/run/finit/cond/run/failure/success" },
-	    "panic":     { "path": "/run/finit/cond/run/failure/failure" }
+	    "startup":   { "path": "/run/finit/cond/usr/startup-config-ok" },
+	    "fail-safe": { "path": "/run/finit/cond/usr/failure-config-ok" },
+	    "panic":     { "path": "/run/finit/cond/usr/failure-config-error" }
 	},
 	"udev": {
 	    "power-a": { "subsystem": "power_supply" },


wkz:
When `confd` became its own daemon (in https://github.com/kernelkit/infix/commit/77a7915e1f7782bc2c6e6a2ffee407905d4cc91b), the conditions used by IITO where also removed.

Currently, a condition is set when startup is applied:
https://github.com/kernelkit/infix/blob/279fc868a11228d5b932d54c74744ca0b11a06fc/src/confd/src/main.c#L844

But there are no corresponding conditions for the failure cases. I suggest we add those and then update the IITO configuration to match.

wkz (changes requested):
Let's fix the root cause so that it (1) fixes all boards and (2) also fixes the error cases when startup fails to apply cleanly.

pjator:
Fixed per your suggestion. confd now sets run/startup/success after a clean bootstrap and run/failure/success/run/failure/failure in handle_startup_failure() (fail-secure + panic paths). The board-specific workaround is dropped. Verified on hardware: LED lights on boot, eMMC visible.

src/confd/src/main.c
```diff
@@ -49,6 +49,31 @@
...
+ * Set a finit condition, e.g. "run/startup/success", creating parent
+ * directories as needed.  Used to signal IITO (and finit services) about
+ * the outcome of the bootstrap/startup sequence.
+ */
+static void set_finit_cond(const char *cond)
+{
+	char path[128];
+	char *p;
+
+	snprintf(path, sizeof(path), "/run/finit/cond/%s", cond);
```
wkz:
```suggestion
	snprintf(path, sizeof(path), "/run/finit/cond/usr/%s", cond);
```
I understand the impulse to use the old condition names from IITO's config. However, the `run/` namespace is reserved by Finit to signal the status of `run` jobs under its control. The `usr/` namespace is allocated for general use by the system.

src/confd/src/main.c
```diff
@@ -49,6 +49,31 @@
...
+
+	/* mkdir -p parent directories (e.g. /run/finit/cond/run/startup) */
+	for (p = path + strlen("/run/finit/cond/"); *p; p++) {
+		if (*p == '/') {
+			*p = '\0';
+			mkdir(path, 0755);
+			*p = '/';
+		}
+	}
```
wkz:
```suggestion
```
Finit guarantees that `/run/finit/cond/usr` exists (via a `tmpfiles` [snippet](https://github.com/finit-project/finit/blob/master/tmpfiles.d/finit.conf)). So with the suggested change above, together with limiting condition names to a single level, we can safely skip this.

src/confd/src/main.c
```diff
@@ -843,6 +871,9 @@ int main(int argc, char **argv)
 	/* Signal that bootstrap is complete (dbus, resolvconf depend on this) */
 	symlink("/run/finit/cond/reconf", "/run/finit/cond/usr/bootstrap");
 
+	/* Signal IITO that startup applied cleanly (status LEDs) */
+	set_finit_cond("run/startup/success");
+
```
wkz:
```suggestion
	set_finit_cond("bootstrap");
```
Let's move the existing condition to use the new helper, but I don't think we can be sure that `startup` has cleanly applied at this point. I have marked that point in a separate comment.

src/confd/src/main.c
wkz, line 613:
```suggestion
		set_finit_cond("startup-config-ok");
		return 0;
```
I believe this is the point where we want to signal a successful application of `startup-config`

wkz, line 526:
```suggestion
	set_finit_cond("startup-config-error");
```
Whatever happens beyond this point, I think we should leave a marker that we could not apply startup-config.

src/confd/src/main.c
```diff
@@ -512,6 +537,7 @@ static void handle_startup_failure(sr_session_ctx_t *sess, const char *failure_p
 		ERROR("sr_copy_config(factory-default) failed: %s", sr_strerror(r));
 		/* Nuclear option: wipe everything */
 		systemf("rm -f /etc/sysrepo/data/*startup* /etc/sysrepo/data/*running* /dev/shm/sr_*");
+		set_finit_cond("run/failure/failure");
 		return;
```
wkz, line 535: @troglobit: Should we not emit the `CRITICAL ERROR: ...` in this scenario as well? The system is still without any defined config in this case, no?

src/confd/src/main.c
```diff
@@ -521,11 +547,13 @@ static void handle_startup_failure(sr_session_ctx_t *sess, const char *failure_p
 			ERROR("Failed loading failure-config, aborting!");
 			banner_append("CRITICAL ERROR: Logins are disabled, no credentials available");
 			systemf("initctl -nbq runlevel 9");
+			set_finit_cond("run/failure/failure");
```
wkz:
```suggestion
			set_finit_cond("failure-config-error");
```
Depending on @troglobit's input, this path may have to be merged with the earlier error path. But I think at think point we want to signal "we tried, but failed, to apply `failure-config`.

src/confd/src/main.c
```diff
@@ -521,11 +547,13 @@ static void handle_startup_failure(sr_session_ctx_t *sess, const char *failure_p
...
 			return;
 		}
 	}
 
 	banner_append("ERROR: Corrupt startup-config, system has reverted to default login credentials");
+	set_finit_cond("run/failure/success");
```
wkz:
```suggestion
	set_finit_cond("failure-config-ok");
```
For symmetry with the other condition names.

wkz (dismissed):
Thanks for taking another stab at this! Sorry about the slew of comments 😅 - you picked a pretty core piece of code for your first contribution.

In addition to these changes, I think `handle_startup_failure()` might need some additional changes. Let's wait and see what @troglobit thinks.

Also, the IITO config needs to be updated to reference the new condition names.

pjator:
@wkz Thanks for the thorough review — it's really helpful, and I agree with the direction (usr/ namespace, single-level condition names, and updating the IITO config accordingly).

I'll wait for @troglobit's input before pushing an updated version, so we can address everything in one pass and match the structure you'd prefer.

I'll follow up once he's had a chance to weigh in.

src/confd/src/main.c
troglobit: Agreed.

src/confd/src/main.c
troglobit: Agreed.

src/confd/src/main.c
troglobit: This is good, agreed.

src/confd/src/main.c
troglobit: Agreed.  The more feedback we can provide the user with in these cases the better.  Possibly those cases could also be consolidated better than what I managed in my refactor.

src/confd/src/main.c
troglobit: Yup, agreed. Just a heads up on the grammar there, I've tried to consistently apply the progressive form of the noun in error and status messages, throughout.  So something like, `CRIT("failed applying failure-config, error %d: %s");` or something.

src/confd/src/main.c
troglobit: Agreed.

troglobit:
Impressive going into the deepest and darkest places, which admittedly needs better lighting.  I agree in general with @wkz, only had a minor comment to add.

pjator:
@wkz @troglobit Thanks for the thorough review! @troglobit — the final frontier admittedly doesn't come with streetlights. 🖖

All feedback addressed and force-pushed (a7c95cd3):

- Conditions now live in the `usr/` namespace (`/run/finit/cond/usr/<name>`), single-level names, and the mkdir loop is gone (Finit guarantees the directory exists)
- confd now signals: `usr/bootstrap`, `usr/startup-config-ok`, `usr/startup-config-error`, `usr/failure-config-ok`, `usr/failure-config-error`
- `startup-config-ok` is only set after startup-config applied cleanly; `startup-config-error` marks the fail-secure path
- Failure paths now also emit a CRITICAL ERROR banner, with messages in progressive form per your convention
- Both iitod.json files (common + nanopi-r2s) reference the new condition names

Build verified (aarch64), hardware test planned for this evening when I get home from work.

troglobit (approved):
Awesome, lgtm!  

Update: @wkz will do some testing on actual HW to verify.

pjator:
Hardware test: IITO conditions verified on physical device

Tested on a NanoPi R2S Plus (32 GB eMMC variant) with the full chain: main + #1586 + #1588 + #1589 + this PR. Built clean from scratch, installed via RAUC (secondary slot), boot status good.

Results:

    usr/startup-config-ok is set by confd after a successful startup-config apply — verified in /run/finit/cond/usr/
    Status LEDs behave as designed: LAN on (startup-config-ok → green:lan), WAN on when link is up (wan-up), SYS off in steady state
    The same conditions are set in test mode (test-config path) — the condition signalling is transport-independent of startup/test config
    Reboot survival: device reboots back into the same slot, boot status: good
    WebUI serves over HTTPS with a generated self-signed cert; IPv4 and IPv6 networking verified

Caveat: the test unit failed at the end of the session. A power-rail component — identified as the 3.3 V LDO (U13) on the USB-C power path per the FriendlyELEC schematic — released smoke and the board no longer powers up. Notably, the unit was powered by a 100 W USB-PD charger; a PD negotiation to a higher voltage (9/12 V) on a circuit designed for 5 V would exceed the LDO's rating and is the most plausible cause (and my mistake). We found no mechanism by which the software changes could cause a power-rail failure, but we cannot entirely rule out that the code or test procedure played a part. The conditions/LED behaviour above was observed before the unit failed.

All software-side validation (boot, conditions, LEDs, WebUI, networking) is complete and passes. Replacement parts are ordered and I will attempt to repair the device, but I cannot provide Infamy test results since the device died while I was running those tests overnight.

Image of component that is smoking. 

image

image

troglobit:
> Hardware test: IITO conditions verified on physical device
> 
> Tested on a NanoPi R2S Plus (32 GB eMMC variant) with the full chain: main + #1586 + #1588 + #1589 + this PR. Built clean from scratch, installed via RAUC (secondary slot), boot status good.

Thanks for verifying!

> Caveat: the test unit failed at the end of the session. A power-rail component — identified as the 3.3 V LDO (U13) on the USB-C power path per the FriendlyELEC schematic — released smoke and the board no longer powers up. Notably, the unit was powered by a 100 W USB-PD charger; a PD negotiation to a higher voltage (9/12 V) on a circuit designed for 5 V would exceed the LDO's rating and is the most plausible cause (and my mistake). We found no mechanism by which the software changes could cause a power-rail failure, but we cannot entirely rule out that the code or test procedure played a part. The conditions/LED behaviour above was observed before the unit failed.

Ouch! :disappointed:

pjator:
**Hardware test NanoPi R2S Plus (eMMC): PASS — no regressions observed**

Tested on NanoPi R2S Plus (eMMC) with the full chain: `main` + #1586 + #1588 + #1589 + this PR (clean build, `v26.06.0-53-g0e1fbc38b`).

A note on the test unit: as commented above it suffered a hardware failure on the power side during my initial testing (the M.2/WiFi regulator, **U13 / SY8089A**, blew up). After some soldering-iron magic I removed U13 — it only supplies the M.2 expansion slot, which is not in use — and the board came back to life, which allowed me to complete the Infamy tests below.

**Infamy suite (test mode, RESTCONF):**
- meta: reproducible, bootorder, check-version — **pass (3/3)**
- system: hostname, add/delete user, admin user, NACM, timezone (name + UTC offset), schedule-reboot — **pass (7/7)**
- hardware: watchdog (hard lockup → trip → reboot) — **pass (6/6)**
- system: upgrade (RAUC install + reboot + boot order) — **pass** (post-verify step fails on test-mode/topology interaction)
- NTP, GPS — **skip** (topology/hardware not applicable to R2S)

**Manual verification (normal operation):**
- factory boot (startup-config cleared → factory config applied, hostname `r2s-97-09-58`, startup-config regenerated) — **pass**
- NTP client synced (chronyd, stratum 3) — **pass**
- IPv4/IPv6 networking — **pass**
- WebUI + generated cert — **pass**
- status LEDs via `usr/startup-config-ok` — **pass** (this PR's change)

**Not passed (pre-existing, not caused by this PR):**
- USB tests fail because the device reports a duplicate hardware component `soc` in `ietf-hardware:hardware` (temperature sensor double-reported via hwmon + thermal zone). I will follow up with a separate PR for the hardware data issue, I think I understand why but want to look at it a little bit more before I come back separately on that.

On a personal note: the Infamy tests are not the easiest to set up and run on physical hardware for a beginner, and I have not managed to produce a `result.json` yet. I hope to include that in future test runs once I have learned the ropes a bit better.

troglobit:
Let's merge this, no more blockers!

PR #1588, opened 2026-08-14
infamy: fix NameError in Topology.get_password (qstrip -> _qstrip)

pjator:
## Description

Fix a `NameError` in the Infamy test harness. `Topology.get_password()`
called `qstrip()` but the helper is defined as `_qstrip()` — so any
physical topology that sets a per-node `password` attribute (or any
value needing quote-stripping) crashed during `env.attach()`.

Found while running the hardware regression suite against a physical
device. Worked around locally, this is the proper upstream fix.

## Checklist

Tick *relevant* boxes, this PR is-a or has-a:

- [x] Bugfix
  - [x] Regression tests
  - [ ] ChangeLog updates (for next release)
- [ ] Feature
  - [ ] YANG model change => revision updated?
  - [ ] Regression tests added?
  - [ ] ChangeLog updates (for next release)
  - [ ] Documentation added?
- [ ] Test changes
  - [ ] Checked in changed Readme.adoc (make test-spec)
  - [ ] Added new test to group Readme.adoc and yaml file
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (please detail in commit messages)
- [ ] Build related changes
- [ ] Documentation content changes
  - [ ] ChangeLog updated (for major changes)
- [ ] Other (please describe):

Disclaimer: Author: Peter Woxblom. Drafting and implementation assisted
by Jarvis, an AI agent (Hermes on DeepSeek V4 Flash backend). Reviewed
and tested on hardware by the author.

PR #1589, opened 2026-08-14
kernel: nanopi-r2s: add rk3328-nanopi-r2s-plus device tree (eMMC support)

pjator:
## Description

Add support for the NanoPi R2S Plus, which adds an on-board 32 GB eMMC
to the plain R2S. The SoC, DRAM and network layout are identical to the
R2S — the only difference is the eMMC, so this is a superset variant
rather than a separate board.

Changes:

- **Device tree:** add `rk3328-nanopi-r2s-plus.dts`, wrapping the
  kernel's `rk3328-nanopi-r2s-plus` (which builds on the shared
  `rk3328-nanopi-r2s.dtsi`), and list both `.dtbs` in `dts/Makefile`,
  following the bpi-r3 pattern. The wrapper overrides the top-level
  compatible to `friendlyarm,nanopi-r2s`, so the Plus shares the R2S
  product configuration — no separate product directory is needed.

- **U-Boot DT selection:** the R2S had no `fdtfile`/`ixvariant`
  mechanism, so U-Boot always loaded the plain r2s DT and the kernel
  never saw the eMMC. Add `ixvariant` (bpi-r4 pattern) that selects
  the plus DT when `mmc0` (eMMC) is present, falling back to the r2s
  DT otherwise.

- **ChangeLog:** entry under v26.08.0.

## Regression Testing (Infamy, physical hardware)

Ran against the NanoPi R2S Plus in test mode
(`touch /mnt/aux/test-mode && reboot`), using the official Infamy
test suite with a single-DUT topology (controller host + target):

- `case/hardware/watchdog`: **6/6 pass** — watchdog tripped on an
  injected hard lockup (all cores, IRQs disabled) and rebooted the
  system back to a working state
- `case/system` (hostname, add/delete user, admin user, NACM,
  timezone x2, schedule reboot): **7/7 pass**
- `case/system/factory_config`: skipped in test mode by design (test
  mode always boots the test config, never the factory config)
- `case/system/upgrade`: not run (requires a data port not wired in
  this lab setup)

Found and worked around an upstream bug in the test harness while
running these tests (`Topology.get_password()` `qstrip` NameError) —
fixed in a separate PR (#1588).

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [x] Feature
  - [ ] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes

Disclaimer: Author: Peter Woxblom. Drafting and implementation assisted
by Jarvis, an AI agent (Hermes on DeepSeek V4 Flash backend). Reviewed
and tested on hardware by the author.

### DIFF
--- a/board/aarch64/friendlyarm-nanopi-r2s/dts/Makefile
+++ b/board/aarch64/friendlyarm-nanopi-r2s/dts/Makefile
@@ -1,1 +1,1 @@
-dtb-y += rockchip/rk3328-nanopi-r2s.dtb
+dtb-y += rockchip/rk3328-nanopi-r2s.dtb rockchip/rk3328-nanopi-r2s-plus.dtb
--- a/board/aarch64/friendlyarm-nanopi-r2s/dts/rockchip/rk3328-nanopi-r2s-plus.dts
+++ b/board/aarch64/friendlyarm-nanopi-r2s/dts/rockchip/rk3328-nanopi-r2s-plus.dts
@@ -1,0 +1,20 @@
+#include <arm64/rockchip/rk3328-nanopi-r2s-plus.dts>
+
+#include "infix.dtsi"
+
+/ {
+	compatible = "friendlyarm,nanopi-r2s", "rockchip,rk3328";
+};
+
+&rtl8153 {
+	/*
+	 * Fix port number: device is on port 1, not port 2.
+	 */
+	reg = <1>;
+};
+
+&{/chosen/infix} {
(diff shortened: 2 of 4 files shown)


troglobit (approved):
lgtm

I really need to give my R2S Plus a go once more. Never managed to get it to boot properly unless I hold the mask button during boot. Unfortunately it's likely a lemon from AliExpress ...

PR #1562, opened 2026-07-02
Add support for unattended updates

saba8814:
This commit introduces unattended-update cron consumer. Unattended updates are configurable to perform update autonomously or to wait for admin to reboot. Unattended upadate reuses UPDATE URL from check-update and is configurable under system/software yang. Cron consumer downloads release tar and extracts .pkg(used by rauc) under /var/lib/misc/unattended-update. PR also introduces migration script due to yang consolidation in relation to update url.

Resolves: #1301

## Checklist

Tick *relevant* boxes, this PR is-a or has-a:

- [ ] Bugfix
  - [ ] Regression tests
  - [ ] ChangeLog updates (for next release)
- [x] Feature
  - [x] YANG model change => revision updated?
  - [ ] Regression tests added?
  - [x] ChangeLog updates (for next release)
  - [ ] Documentation added?
- [ ] Test changes
  - [ ] Checked in changed Readme.adoc (make test-spec)
  - [ ] Added new test to group Readme.adoc and yaml file
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (please detail in commit messages)
- [ ] Build related changes
- [ ] Documentation content changes
  - [ ] ChangeLog updated (for major changes)
- [ ] Other (please describe):

PR #1629, opened 2026-09-14
Add support for unattended updates

### LINKED ISSUE #1301: Add support for unattended upgrades

saba8814:
This commit introduces unattended-update cron consumer. Unattended updates are configurable to perform update autonomously or to wait for admin to reboot. Unattended upadate reuses UPDATE URL(pointing to Atom/RSS feed) from check-update and is configurable under system/software yang. Cron consumer runs rauc install .pkg thus HTTP/s range shall be supported on the web server. Also docs are included and migration script due to changes in update url. Original PR: #1562
Resolves: #1301

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [x] Feature
  - [ ] Regression tests added?
  - [x] Documentation added?
- [ ] Test changes
- [ ] Build related changes
  - [x] ChangeLog updated (for major changes)


doc/schedule.md
```diff
@@ -0,0 +1,149 @@
+# Scheduling
+
+Some Infix features run on a recurring calendar instead of on demand.
```
troglobit: We try to avoid hard-coding the name *Infix* in documentation, this primarily to avoid painful renames when brand labeling Infix for customers.

Usually the text reads better when keeping this in mind.  Here an alternative text could instead say; "Some system functions run periodically instead of on demand.  One such type of recurrence is scheduled activities, like checking for new software updates."

Notice how this first sentence is not just easier to read, but also introduces *the why* of the entire section.  This is a good thing to remember even in spoken languages, *"start with the why"*, and it translates well to written form as well.

Please have a look at and try to apply the rule and the recommendation to the rest of this document.

doc/upgrade.md
```diff
@@ -194,6 +194,218 @@ now the preferred boot source.
 To upgrade the remaining partition (`primary`), run the `upgrade URL`
 command again, and (optionally) reboot.
 
+## Unattended Updates
+
+The upgrade above is operator-driven: you pick a bundle, run `upgrade`,
+and reboot.  Infix can also perform that same install on its own, on a
+recurring [schedule][6].
+
```
troglobit, line 202: Same here.  Alternative; "This is a function the system can perform on its own, using a [schedule][6]."

doc/upgrade.md
```diff
@@ -194,6 +194,218 @@ now the preferred boot source.
...
+- **One bundle per platform**, named `<image-id>-<tag>.pkg`.  A device
+  looks only for its own `IMAGE_ID`, so one feed can serve several
+  platforms
+- **HTTP range requests.**  RAUC streams the bundle instead of downloading
+  it whole, so a server that ignores `Range` fails the install.  BusyBox
+  `httpd` and nginx both work; Python's `http.server` does not
+
+Infix uses the `/releases/tag/<tag>` URL only to derive the base and as a
+human-readable link in the update-check notification.  The page itself
+does not have to exist.
```
troglobit: Same here, and here's another general writing recommendation: Simplify.  To solve this one, we can simply drop the leading two words: "The `/releases/tag/<tag>` URL only ..."

doc/upgrade.md
```diff
@@ -194,6 +194,218 @@ now the preferred boot source.
...
+
+A single-instance lock means occurrences never overlap: if an install is
+still running when the next one fires, the new occurrence is skipped.
+
+> [!IMPORTANT]
+> Rollback safety is identical to a manual `upgrade`: the previously
+> running image remains on the other partition, and the bootloader falls
+> back to it if the new image does not boot.  Infix does no health check
+> on the new image beyond that, so see the caution under
+> [Upgrading](#upgrading) about upgrading only one partition at a time.
```
troglobit: Hmm, we don't really have full rollback support yet.  Suggest we drop this entirely, or reword it heavily (Infix name again).  But let's consult @jovatn on this first!

src/confd/yang/confd/infix-system-software.yang
```diff
@@ -93,13 +100,40 @@ submodule infix-system-software {
...
+           the latest version; each entry is expected to link to its release
+           page as '<base>/releases/tag/<tag>', from which the tag is read.
+
+           The feed carries no asset list, so the per-platform bundle is
+           fetched by convention from
+           '<base>/releases/download/<tag>/<image-id>-<tag>.pkg', and
+           installed by streaming it straight from that URL.
+
+           Override for a fork or a customer-specific channel.";
+      }
```
troglobit, line 118: @mattiaswal I'd like to suggest we do not encode the github url to kernelkit/infix in a YANG model default.  Imo this should be injected in the generated factory-config and board-specific factory-config.cfg files instead.

mkdocs.yml
troglobit, line 62: When we have the option, alphabetical ordering is to be preferred.  I broke that rule by placing Access Control further down ... I could easily be convinced to move that up when you move Scheduling below Keystore.

troglobit (changes requested):
Nice work on this!

Lots of minor things still to change and discuss with other team members.

doc/schedule.md
saba8814: Great i will add it to my AGENT.md in future.

mkdocs.yml
saba8814: I will move yours aswell to no worries :smile:

doc/upgrade.md
jovatn: I can agree to dropping the text.
But if we wish to indicate that things can go wrong (while waiting for full rollback to be implemented), we could perhaps rephrase it like this?:
```
[!Caution]
Unattended upgrade does no additional checks than done by a manual `upgrade`: the previously ...
```

troglobit: That's good, thanks @jovatn, let's go with that!

src/confd/src/core.c
```diff
@@ -563,6 +563,56 @@ static confd_dependency_t dep_radio_components(struct lyd_node **diff, struct ly
 	return result;
 }
 
+static confd_dependency_t dep_schedule_consumers(struct lyd_node **diff, struct lyd_node *config)
+{
+	confd_dependency_t result = CONFD_DEP_DONE;
```
mattiaswal, line 568: Still to complex, talked with claude, he suggested:

```
        struct ly_set *users;
        uint32_t i;

        if (!lydx_get_xpathf(*diff, "/ietf-system:system"))
                return CONFD_DEP_DONE;

        /* Any feature pointing a schedule-ref at a schedule, per the
         * infix-schedule contract.  Re-adding an unchanged schedule is a
         * no-op in add_dependencies(). */
        users = lydx_find_xpathf(config, "/ietf-system:system//*[schedule]");
        for (i = 0; users && i < users->count; i++) {
                struct lyd_node *sched = lydx_get_child(users->dnodes[i], "schedule");
                const char *name = sched ? lyd_get_value(sched) : NULL;
                char xpath[256];

                if (!name)
                        continue;

                snprintf(xpath, sizeof(xpath),
                         "/ietf-system:system/infix-schedule:schedules/schedule[name='%s']", name);
                result = add_dependencies(diff, xpath, name)
```

mattiaswal (changes requested):

src/confd/yang/confd/infix-system-software.yang
saba8814: Any decision on this one since I am currently fixing rest of the comments?

mattiaswal: Yes, remove it from as default in the model, it's right place is the factory-config (but not now), and it should be set as mandadtory in the model.

src/confd/src/core.c
mattiaswal: Always have as a habit to ask for simpler solutions, claude has a bad habit of overcompicate stuff.

saba8814:
2. Re‑installs the same bundle on every occurrence until reboot (medium, robustness)
With reboot manual, running VERSION never changes after a successful install, so each schedule tick downloads and re‑flashes the identical bundle to the inactive slot. Dev builds (IS_RELEASE=false) do this unconditionally. That is needless flash wear and bandwidth, and it makes the lock the only thing preventing overlap. Check the inactive slot's version via rauc status --output-format=json (or remember the last installed tag) and skip when it already matches LATEST_TAG.

mattiaswal (approved):

troglobit (approved):

PR #1632, opened 2026-09-14
Support collect

mattiaswal:
The main goal for this PR is to fix a security issue in the WebUI that support collect did not honor NACM, any user could get the support bunde (including all passwords). Now the support bundle honor NACM and also reduct all passwords, they are of no use for when debugging issues.

## Checklist

- [x] Bugfix
  - [x] Regression tests
- [x] Feature
  - [x] Regression tests added?
  - [x] Documentation added?
- [x] Test changes
- [ ] Build related changes


doc/support.md
```diff
@@ -60,6 +63,77 @@ It shows what was collected and what failed. Use `support clean` to
...
+  /var/lib/support/support-host-2026-09-11T13:05:42+02:00.tar.gz
+  $ scp admin@host:/var/lib/support/support-host-2026-09-11T13:05:42+02:00.tar.gz .
+  ```
+- A collection that fails keeps its log in `/var/lib/support`. The error
+  message says why and names the log.
+- Each call removes what earlier calls left behind once it is a week
+  old, the same as `support clean` does.
+
+From a shell on the device, use the `support` command rather than the
+RPC. A base64 blob on your terminal is of no use to anyone.
```
troglobit: This text is waaay too long, intricate, and talks too much about implementation details.  Please simplify and focus on the core bits an end user would otherwise trip over.

Also, the SSH alternative is already documented so we should be able to just add an intra-link reference instead of repeating an example -- if the existing examples are incomplete, we should update them instead.

troglobit (changes requested):
I must admit, I was a bit skeptical at first about the premise of doing this over an RPC -- and I'm still worried about the 60 sec. timeout coming to bite us -- but it turned out much better than I feared so it's a go from me, except for the documentation bits.

Nice work on all the side issues as well! 🌟

mattiaswal:
> I must admit, I was a bit skeptical at first about the premise of doing this over an RPC -- and I'm still worried about the 60 sec. timeout coming to bite us -- but it turned out much better than I feared so it's a go from me, except for the documentation bits.
> 
> Nice work on all the side issues as well! 🌟

Me too, but it is mainly when doing it over NETCONF/RESTCONF, when you do it over the web, the connection is to localhost :)

doc/support.md
mattiaswal: Agree.

troglobit:
> > I must admit, I was a bit skeptical at first about the premise of doing this over an RPC -- and I'm still worried about the 60 sec. timeout coming to bite us -- but it turned out much better than I feared so it's a go from me, except for the documentation bits.
> 
> > 
> 
> > Nice work on all the side issues as well! 🌟
> 
> 
> 
> Me too, but it is mainly when doing it over NETCONF/RESTCONF, when you do it over the web, the connection is to localhost :)

Really looking forward to your yangerd work here. Will speed up support tool a lot

mattiaswal:
> > > I must admit, I was a bit skeptical at first about the premise of doing this over an RPC -- and I'm still worried about the 60 sec. timeout coming to bite us -- but it turned out much better than I feared so it's a go from me, except for the documentation bits.
> > 
> > 
> > > 
> > 
> > 
> > > Nice work on all the side issues as well! 🌟
> > 
> > 
> > Me too, but it is mainly when doing it over NETCONF/RESTCONF, when you do it over the web, the connection is to localhost :)
> 
> Really looking forward to your yangerd work here. Will speed up support tool a lot

Yupp, any century now.

troglobit (approved):

Issue #514, opened 2024-07-02
Full ietf-interfaces YANG support (advanced)


troglobit:
### Tasks
- [x] [ietf-ip: ip:neighbor](https://github.com/kernelkit/infix/issues/819)
- [x] ietf-ip: ipv4+ipv6 mtu support
- [ ] ietf-intefaces: last-change
- [ ] ietf-intefaces: higher-layer-if
- [ ] ietf-intefaces: lower-layer-if
- [x] #152 
- [x] #515

PR #1635, opened 2026-09-14
Finalize ietf-interfaces.yang: last-change, higher-layer-if, lower-layer-if

troglobit:
## Description

Last interface change

Completes #514 with the remaining ietf-interfaces.yang status leaves. Both CLI and WebUI support.

- higher-layer-if and lower-layer-if come from the kernel's
  upper_*/lower_* links in sysfs, covering bridge ports, LAG
  members, VLAN parents and any other netdev adjacency.
- last-change is tracked by statd over rtnetlink in an in-memory
  map keyed by interface name.  Interfaces whose state predates
  statd have no last-change, per RFC 8343.
- `show interface <name>` prints last change and the layer lists.

> [!NOTE]
> As per RFC8343, the `last-change` node is not shown (included) "If the current state was entered prior to the last re-initialization of the local network management subsystem".  I.e., on startup.

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [x] Feature
  - [x] Regression tests added?
  - [ ] Documentation added?
- [ ] Test changes
- [ ] Build related changes


mattiaswal (approved):

jovatn:
@troglobit Great enhancement!!

Some comments:
- I do not see any update to yang among the changed files. I suppose I look at the wrong place?!
- I also note that there are no updates to doc except for changelog, and it's because we do not have any `show interface br0`or `show interface veth0a` examples. Should we add that? (separate issue)
- How is it with link aggregates, does "lower-layer" status apply to that? (suppose not, but I am unsure)

troglobit:
> @troglobit Great enhancement!!

thx :sunglasses: 

> Some comments:
> * I do not see any update to yang among the changed files. I suppose I look at the wrong place?!

These nodes were never deviated, just missing backend code.

> * I also note that there are no updates to doc except for changelog, and it's because we do not have any `show interface br0`or `show interface veth0a` examples. Should we add that? (separate issue)

It's honestly such a minor thing, I think we can skip it for now.  There's a lot of bridge changes coming, and possibly even veth if we do the podman swap, so the docs will be updated over time anyway.

> * How is it with link aggregates, does "lower-layer" status apply to that? (suppose not, but I am unsure)

It does,  the data comes from the kernel's netdev adjacency, and the bonding driver registers exactly that link when a port joins a LAG, the same call the bridge uses for its ports.  So a LAG member shows the LAG under `higher-layer-if`, and the LAG lists its members under `lower-layer-if`.

PR #1643, opened 2026-09-15
Add RPC for injecting messages in the system log

### LINKED ISSUE #1639: Add RPC for injecting log messages to a device
### Description

Test systems, and scripts in general, have no clean way to leave a mark in a device's system log.  Today this means logging in over SSH and running logger(1), which is clumsy on a system with NETCONF/RESTCONF.

### Additional Information

This feature request is primarily for the test servers, but possibly handy for other use-cases as well.

### General Information

troglobit:
## Description

Test systems and scripts have had no clean way to leave a mark in a device's log, short of SSH and `logger(1)`.  This adds an RPC that hands a message to the local syslog daemon, exposing the full RFC 5424 header: severity, app-name, msgid, and structured data.  Defaults follow `logger`, `user.notice`, with app-name defaulting to the calling user.  Messages are time stamped on arrival and follow the configured `/syslog` filtering and forwarding rules, so the device decides where they end up, not the caller.

- **confd:** new `infix-syslog.yang` revision with the `log` RPC
- **cli:** admin-exec `log [severity S] [msgid M] MESSAGE`.
- **infamy:** `target.log()` helper, four tests converted from SSH + `logger`
- **test:** new `syslog/rpc_log` case verifying header fields and defaults

## Checklist

- [ ] Bugfix
  - [ ] Regression tests
- [x] Feature
  - [x] Regression tests added?
  - [x] Documentation added?
- [ ] Test changes
- [ ] Build related changes


mattiaswal (commented):
Security review of the new `log` RPC. Verified against libyang (yanglint) with the exact type definitions from this PR, and against the sysklogd 2.7.2 sources Infix ships. Details inline; summary ranked by severity:

1. **High:** format string injection via `msgid`. libsyslog's `vsyslogp_r()` concatenates `msgid` into the printf format it hands to `vsnprintf()`, and the pattern `[!-~]+` allows `%`. Runs inside confd as root.
2. **Medium:** log line forgery via `structured-data/param/value`. Control characters are not escaped, syslogd does not sanitize SD (only MSG), and `fmt5424` writes it verbatim to RFC 5424 files and remote forwarding.
3. **Medium:** unbounded SD is truncated in libsyslog's 2048-byte buffer, after which syslogd's RFC 5424 parser drops the whole message silently.
4. **Medium (design):** any user with `exec` rights (factory NACM: `exec-default permit`) can forge entries tagged as any daemon, in `auth`/`authpriv`, or at `emergency` severity (walled to all ttys and `/dev/console`). Only `kern` is neutralized by syslogd.
5. **Low:** no rate limiting; libsyslog's `sendto()` is blocking and confd's RPC callback is `SR_SUBSCR_NO_THREAD`.
6. **Low:** default app-name from `sr_session_get_orig_name()` is client-controlled.

What I checked and found fine: the `sd-name` typedef matches RFC 5424 exactly (rejects `=`, `]`, `"`, SP); `sd_build()` length arithmetic is correct and sysrepo validates mandatory/key leaves before the callback; ASCII control characters in `message` are rejected by libyang (C0 except TAB/LF/CR) or neutralized by syslogd (LF to space, others to `^X`); the klish script quoting and the `rpc`/`copy` changes are sound.

Note that `-8` on syslogd disables C1 masking, so UTF-8 encoded C1 controls (U+0085, U+009B), U+2028 and bidi overrides in `message` reach log files and `show log` untouched. Low severity, mentioned for completeness.

src/confd/yang/confd/infix-syslog.yang
```diff
@@ -285,4 +303,85 @@ module infix-syslog {
...
+          pattern '[!-~]+';
+        }
+        description "Originating application name, or tag, RFC 5424 APP-NAME.
+                     Default: name of the calling user.";
+      }
+
+      leaf msgid {
+        type string {
+          length "1..32";
+          pattern '[!-~]+';
```
mattiaswal, line 355: **High: format string injection.** This pattern permits `%`, and sysklogd's `vsyslogp_r()` builds its printf format by concatenating `msgid` into it (sysklogd 2.7.2 `src/syslog.c` L451-452, then `vsnprintf(p, tbuf_left, fmt_cpy, ap)` at L507). A caller can therefore pass `msgid` = `%s%s%s%s%s%s` to crash confd, or `%n...` for a write primitive, and this runs as root. yanglint confirms `%n%n%n%n` is accepted by this type.

Suggest excluding `%` here, e.g. `pattern '[!-$&-~]+';`, and also escaping `%` as `%%` in `rpc_log()` so the C side does not depend on the model. `app-name` is not affected (libsyslog uses `snprintf("%s ", tag)`), and `sd`/`message` are passed as `%s` arguments.

src/confd/src/syslog.c
```diff
@@ -478,3 +480,192 @@ int syslog_change(sr_session_ctx_t *session, struct lyd_node *config, struct lyd
...
+	pri   = log_facility(lydx_get_cattr(in, "facility")) | log_severity(lydx_get_cattr(in, "severity"));
+	msgid = lydx_get_cattr(in, "msgid");
+	tag   = lydx_get_cattr(in, "app-name");
+	if (!tag)
+		tag = log_user(session);
+
+	log.log_tag = tag;
+	sd = sd_build(in);
+	if (sd)
+		syslogp_r(pri, &log, msgid, "%s", "%s", sd, msg);
```
mattiaswal, line 652: See the comment on the `msgid` leaf: `msgid` is used as part of the format string inside `syslogp_r()`, not as a plain argument. Even with the YANG pattern tightened, I would escape `%` to `%%` here (or reject the value) before passing it on. The `"%s"` handling of `sd` and `msg` is correct.

src/confd/yang/confd/infix-syslog.yang
```diff
@@ -285,4 +303,85 @@ module infix-syslog {
...
+          key "name";
+          description "SD-PARAM, name and value pair.";
+
+          leaf name {
+            type sd-name;
+            description "Parameter name.";
+          }
+
+          leaf value {
+            type string;
```
mattiaswal, line 379: **Medium: log line forgery and silent message loss.**

1. `value` is an unrestricted string, and `sd_escape()` only escapes `"`, `\` and `]`. libyang accepts LF, CR, DEL and C1 controls here. syslogd's `parsemsg_rfc5424()` accepts any bytes inside PARAM-VALUE except an unescaped `"`, keeps the SD block raw (`buffer.sd = msg`), and `fmt5424()` writes it verbatim. Only MSG goes through `parsemsg_remove_unsafe_characters()`. So a value containing `\n` produces a fabricated extra log line in every `log-format rfc5424` file and in RFC 5424 remote forwarding.

2. There is no bound on value length or list size, but libsyslog formats the whole packet into `tbuf[2048]`. A long SD is truncated mid-element, syslogd's `FAIL_IF("STRUCTURED-NAME", ...)` then drops the entire message, and the caller gets `SR_ERR_OK`.

Suggest something like `length "1..255"` plus a pattern excluding control characters, and escaping or stripping controls in `sd_escape()` as a belt-and-braces measure.

src/confd/src/syslog.c
```diff
@@ -478,3 +480,192 @@ int syslog_change(sr_session_ctx_t *session, struct lyd_node *config, struct lyd
...
+	}
+
+	if (orig && orig[0])
+		return orig;
+
+	return sr_session_get_user(session);
+}
+
+/* RFC 5424 PARAM-VALUE: escape '"', '\\', and ']' */
+static char *sd_escape(char *ptr, const char *value)
```
mattiaswal, line 582: Consider also handling control characters here (e.g. replace anything `< 0x20` and `0x7f` with a space), since syslogd sanitizes MSG but not SD. See the comment on the `value` leaf for the forgery scenario.

src/confd/yang/confd/infix-syslog.yang
```diff
@@ -285,4 +303,85 @@ module infix-syslog {
...
+    description "Log a message via the local system logger.
+
+                 The message is handed to the system log daemon like any
+                 locally generated message, i.e., it is time stamped on
+                 arrival and subject to the same filtering and forwarding
+                 rules as configured in /syslog.";
+    reference "RFC 5424: The Syslog Protocol";
+
+    input {
+      leaf message {
```
mattiaswal, line 321: FYI on control characters in `message`, since it came up: libyang enforces RFC 7950 string rules, so C0 controls other than TAB/LF/CR, surrogates and noncharacters are rejected at parse time (ESC cannot get in). LF/CR/DEL do get through, but syslogd's `parsemsg_remove_unsafe_characters()` turns LF into a space and the rest into `^X`, so this leaf cannot forge a second log line. Because Infix runs `syslogd -8`, C1 masking is off, so UTF-8 encoded U+0085/U+009B, U+2028 and bidi overrides such as U+202E are logged as-is. Low severity; a pattern like `[^\p{Cc}\p{Cf}\p{Zl}\p{Zp}]+` would close it if wanted.

One more thing: `length "1..2048"` counts characters, while libsyslog's buffer is 2048 bytes for the entire packet including header and SD, so long messages are silently truncated. Worth mentioning in the description or lowering the limit.

src/confd/yang/confd/infix-syslog.yang
```diff
@@ -285,4 +303,85 @@ module infix-syslog {
     description "Not yet supported by underlying daemon.";
     deviate not-supported;
   }
+
+  /*
+   * RPCs
+   */
+
+  rpc log {
```
mattiaswal, line 311: **Medium (design): forgery and authorization.** Callers pick `app-name`, facility and severity freely. That allows entries in `/var/log/auth.log` tagged `sshd` with facility `auth`, or `emergency` messages that syslogd walls to every logged-in tty and to `/dev/console` (factory `*.=emerg *` and `console.conf`). Only `kern` is neutralized, since syslogd remaps LOG_KERN from `/dev/log` unless started with `-k`.

The factory NACM has `exec-default permit`, so operators and any non-guest user can call this. Options: a factory NACM rule limiting `infix-syslog:log` exec to admin (or an explicit operator permit), refusing `auth`/`authpriv`, and/or always adding a fixed marker (e.g. an SD element or msgid prefix) so injected messages stay distinguishable from genuine ones in forwarded logs. At minimum the docs should say that the header fields are caller-supplied and not authenticated.

src/confd/src/syslog.c
```diff
@@ -478,3 +480,192 @@ int syslog_change(sr_session_ctx_t *session, struct lyd_node *config, struct lyd
...
+			*ptr++ = '"';
+		}
+		*ptr++ = ']';
+	}
+	*ptr = 0;
+
+	return sd;
+}
+
+static int rpc_log(sr_session_ctx_t *session, uint32_t sub_id, const char *op_path,
```
mattiaswal, line 629: **Low: no rate limiting.** Each call can write up to 2048 bytes to a typically RAM-backed `/var/log`, rotating away real evidence and flooding remote collectors. Also, libsyslog's `sendto()` on the connected AF_UNIX datagram socket is blocking (no `SOCK_NONBLOCK`), and this callback is registered with `SR_SUBSCR_NO_THREAD`, so a full syslogd receive queue would stall the whole confd event loop. A simple token bucket here would cover both.

src/confd/src/syslog.c
```diff
@@ -478,3 +480,192 @@ int syslog_change(sr_session_ctx_t *session, struct lyd_node *config, struct lyd
...
+
+	return LOG_NOTICE;
+}
+
+/*
+ * The event session runs as confd, the calling user is only known from
+ * the originator: netopeer2 pushes [nc-sid, username], the CLI and the
+ * rpc tool set their originator name to the user.
+ */
+static const char *log_user(sr_session_ctx_t *session)
```
mattiaswal, line 564: **Low:** `sr_session_get_orig_name()` is set by the client itself (`copy.c` does this too), so any sysrepo client can claim any name, and one claiming `netopeer2` without orig data falls through to the literal string `netopeer2` as tag. No privilege is gained since `app-name` is caller-controlled anyway, but the docs should not present the default as an authenticated identity.

mattiaswal (commented):
Duplicate of the review above, posted twice by mistake. Comments removed.

mattiaswal:
Didn't find anything when reviewing but i was scared about something like https://github.com/kernelkit/infix/pull/1643#discussion_r4026036753 so i let claude get a round of this. This i feel is crutial, other it looks fine.

src/confd/yang/confd/infix-syslog.yang
troglobit: I was considering marking the RPC with `nacm:default-deny-all;`, so this is great feedback, thanks!  It's the age-old question of flexibility vs security.

src/confd/yang/confd/infix-syslog.yang
troglobit: Awesome feedback, will look into this asap!

src/confd/src/syslog.c
troglobit: Will fix, thanks!

src/confd/yang/confd/infix-syslog.yang
troglobit: I'll take a deeper look at this too, the formatting rules are strict but allows quite a lot still, so this needs some careful thinking.  Might even spill over as fixes in the sysklogd project, so thank you! :pray:

troglobit:
> Didn't find anything when reviewing but i was scared about something like [#1643 (comment)](https://github.com/kernelkit/infix/pull/1643#discussion_r4026036753) so i let claude get a round of this. This i feel is crutial, other it looks fine.

Agreed, will have a look at it over the weekend, thanks! :bowing_man:

src/confd/yang/confd/infix-syslog.yang
mattiaswal: Guess it can be good, we should really be more strict about access in general and use nacm:default-deny-all;  more.

troglobit: Fixed.

src/confd/yang/confd/infix-syslog.yang
troglobit: Fixed.

src/confd/src/syslog.c
troglobit: Fixed.

src/confd/src/syslog.c
troglobit: Fixed.

troglobit:
Ping @mattiaswal: rebased on latest `main` and `facility` is now removed, so not possible anymore to forge `avahi-daemon`, for example.

mattiaswal:
Verified fixed

- Format string injection (High). The msgid pattern rejects %, confirmed with yanglint on the PR's module. The C side also doubles any % before libsyslog splices it into the printf format. Either fix alone is sufficient.
- SD log line forgery (Medium). The value pattern rejects all Unicode control characters. yanglint rejects LF, CR, TAB, DEL and the C1 U+0085. The C escaper also maps anything below 0x20 to a space as a fallback.
- Silent truncation (Medium). A pre-check computes the full RFC 5424 line before sending and rejects oversize input with an error. I traced the arithmetic against sysklogd 2.7.2. It over-estimates by ten bytes, so it is conservative in the safe direction.
- Authorization (Medium). The RPC now carries default-deny-all. In the factory NACM only the admin permit-all rule matches it, and I confirmed sysrepo applies the extension at step 10 after no rule matches. NETCONF, RESTCONF and the CLI all set the NACM user, so the deny holds on every path. The facility leaf is gone and the code hard-wires user, so auth, authpriv and kern are unreachable.

Also closed: the literal "netopeer2" tag fallback now yields -.

Unchanged, acceptable with admin-only exec

- No rate limiting, and the blocking send to the log socket still runs inside the no-thread callback.
- emergency still walls to all ttys and the console.
- The app-name pattern allows [, ] and :, so an admin can produce an RFC 3164 line that looks like sshd[123]:. The user facility keeps it out of the auth log.
- U+2028 and bidi overrides pass both message and SD values. Cosmetic to viewers only.


Verdict: no open findings above Low. Safe to merge from the security side.

mattiaswal (approved):

